PR #113, opened 2021-12-22
chore: update NPM modules using npm audit fix

paulschreiber:
## Description
update NPM modules using `npm audit fix`.


paulschreiber:
@josebui There's only one alert:`ansi-html`. Looks like it doesn't have a patch yet, so this just updates other dependencies.

josebui (approved):

PR #117, opened 2021-12-23
feat: Auth integration

### LINKED ISSUE #17: User Sign in and Sign up process

josebui:
## Description

- Added integration with auth providers

### Checklist
- [x] New tests added
- [x] Strings are localized
- [x] Verified on mobile
- [x] Verified on desktop

### Related Issues
Fixes #17


src/account/accountService.js
```diff
@@ -0,0 +1,34 @@
...
+  {
+    headers: {
+      Authorization: `Bearer ${getToken()}`,
+      'Content-Type': 'application/json'
+    }
+  }
+)
+  .then(response => {
+    if (_.includes([401, 403], response.status)) {
+      return Promise.reject('UNAUTHENTICATED')
```
paulschreiber, line 30: 401 is unauthorized (no creds); 403 is forbidden (creds but not access to this resource).

src/common/components/PageLoader.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { Backdrop, CircularProgress } from '@mui/material'
+
+const PageLoader = () => (
+  <Backdrop
+    sx={{ color: 'white', zIndex: theme => theme.zIndex.drawer + 1 }}
```
paulschreiber, line 6: Would white be the default backdrop?

src/navigation/Routes.js
```diff
@@ -13,23 +9,41 @@ import LandscapeList from 'landscape/components/LandscapeList'
...
-import AccountProfile from 'auth/components/AccountProfile'
+import AccountLogin from 'account/components/AccountLogin'
+import AccountProfile from 'account/components/AccountProfile'
+import RequireAuth from 'account/components/RequireAuth'
+
+const path = (path, Component, auth = true) => ({
+  path, Component, auth
+})
+
+const paths = [
```
paulschreiber, line 20: 👍

src/state/utils.js
```diff
@@ -13,6 +15,15 @@ export const createAsyncThunk = (name, action, onSuccessMessage, customErrorMess
       }
       return result
     } catch (error) {
+      // Handle auth error
+      if (error === 'UNAUTHENTICATED') {
```
paulschreiber, line 19: See notes above (L30).

src/terrasoBackend/api.js
```diff
@@ -9,18 +10,26 @@ const handleGraphQLError = data => {
...
 export const request = async (query, variables) => {
-  const response = await fetch(TERRASO_API_URL, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json'
-    },
-    body: JSON.stringify({ query, variables })
-  })
+  const response = await fetch(
+    new URL('/graphql/', TERRASO_API_URL).href,
```
paulschreiber, line 14: Q: Should `/graphql/` be in a constant somewhere?

src/account/accountService.js
josebui: For me in this particular case (getting the logged user data) 403 and 401 mean that something is wrong with your "session" that is why here I consider both to force the relogin

src/common/components/PageLoader.js
josebui: @paulschreiber Yes, for this one I just extracted all the duplciated code for the Backdrop, nothing is changing

src/state/utils.js
josebui: @paulschreiber Can you add a little more context for this? Do you think we should handle other codes? or just 401?

src/terrasoBackend/api.js
josebui: It will only be used here, it can be a constant in this same file, but I don't see us changing this ever so maybe it is better to have it directly here, but let me know what do you think

src/common/components/PageLoader.js
paulschreiber: Would it work without specifying the color?

src/state/utils.js
paulschreiber: If we get a 401, we want to bounce them to a login screen. If we get a 403, we should display an error.

src/terrasoBackend/api.js
paulschreiber: Would still be good to put in a config section somewhere.

src/common/components/PageLoader.js
josebui: @paulschreiber It works, the loader jsut displays black, I can simplify it to that since there is no erquirement for that white, let me know what do you think

paulschreiber (approved):

Issue #115, opened 2021-12-22
Form validation: allow URLs without protocol


paulschreiber:
## Description
Currently, we accept `https://www.example.com` but not `www.example.com`.

We should accept both, and automatically add the URL scheme if the user omits it.

ssfrancis:
How will we know if it should be http:// or https://?  Can we / will we test programmatically?

paulschreiber:
99% of sites are https. If someone needs `http`, they will have to enter it manually.

Hanal-code:
Change the input field not to require the users to enter 'http://www...' The web address bar starts with the name nowadays.  e.g. 'terrso.org' in the address bar.
![image](https://user-images.githubusercontent.com/92342777/147967759-17b51917-f8a8-43eb-b441-d8a1e4da7f7c.png)

PR #142, opened 2022-01-05
fix: GroupForm URL field validation

cacarrara:
This change updates the URL field validation on `GroupForm` to allow users
inform URLs without the protocol (HTTPS). To make it work, it was added
a transform function to add the protocol value if needed before calling
the field validation and sending the data to the API.

This fix:
- #115

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,10 @@
     "serve": "^13.0.2"
   },
   "jest": {
+    "collectCoverageFrom" : [
+      "src/**/*.js",
+      "!src/terrasoBackend/*.js"
+    ],
     "coverageThreshold": {
       "global": {
         "branches": 80,
--- a/src/group/components/GroupForm.js
+++ b/src/group/components/GroupForm.js
@@ -14,13 +14,21 @@ import { fetchGroupForm, saveGroup, setFormNewValues, resetFormSuccess } from 'g
 import Form from 'forms/components/Form'
 import PageLoader from 'common/components/PageLoader'
 
+const transformURL = (url) => {
+  if (url === '' || url.startsWith('http:') || url.startsWith('https:')) {
+    return url
+  }
+
+  return `https://${url}`
+}
+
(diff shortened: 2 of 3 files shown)


josebui:
@cacarrara Looks good, then only thing that I don't see is the value that will go to the API changing, with this it seems that still the URL will be stored without the protocol, is that part of a different issue? Or should we add the same logic when rendering the links?

cacarrara:
@josebui this is the line changing the value that goes to the API:

https://github.com/techmatters/terraso-web-client/pull/142/files#diff-7d965c97921b9630cfe254b57bf6a2f06215ba5f4ca739fae4e6b87ced4ab46aR22

This is being covered in the added test.

cacarrara:
Also, the API doesn't allow URL without protocol. So, if you run this branch locally, you can see that the request is being done successfully, even without informing the protocol.

josebui (approved):
Ohh, I missed that, so yup is not only validating it is transforming and that is why you call it transform. Looks good

cacarrara:
@josebui do you have any recommendation how to make coverage pass? :thinking:

josebui:
@cacarrara Now, until we fix it there is not much we can do. But the coverage looks good:
```
------------------|---------|----------|---------|---------|-------------------
File              | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
------------------|---------|----------|---------|---------|-------------------
All files         |   69.49 |    68.42 |   72.22 |   68.51 |                   
 group/components |     100 |      100 |     100 |     100 |                   
  GroupForm.js    |     100 |      100 |     100 |     100 |                   
 terrasoBackend   |      10 |        0 |       0 |   10.52 |                   
  api.js          |      10 |        0 |       0 |   10.52 | 8-10,14-49        
 tests            |     100 |      100 |     100 |     100 |                   
  utils.js        |     100 |      100 |     100 |     100 |                   
------------------|---------|----------|---------|---------|-------------------
```

PR #147, opened 2022-01-10
feat: Tables URL params

### LINKED ISSUE #136: Include pagination and sort order in URL
## Description
Currently, clicking on the next page does not update the URL. This prevents sharing of the URL and breaks the back button.

This applies to landscapes and groups.

In the future, we will also need to include search term and page size.

## Steps To Reproduce
Steps to reproduce the behavior. Describe what you typed, clicked, tapped, swiped, saw and heard.
1. Go to URL https://app.staging.terraso.org/landscapes
2. Click on > to go to next page.

## Expected behavior
URL changes to https://app.staging.terraso.org/landscapes/?page=2 or https://app.staging.terraso.org/landscapes/page/2/ or https://app.staging.terraso.org/landscapes/2/. Sort order should use a query parameter (i.e. `?sort=name&order=asc`)

## Actual behavior
URL is unchanged.

josebui:
## Description

- Added URL query params when table parameters change

### Checklist
- [ ] New tests added
- [ ] Strings are localized
- [ ] Verified on mobile
- [ ] Verified on desktop

### Related Issues
Fixes #136 

### Verification steps

Screenshot:
Screen Shot 2022-01-10 at 12 16 01 PM

### DIFF
--- a/src/common/components/Table.js
+++ b/src/common/components/Table.js
@@ -1,18 +1,74 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
+import _ from 'lodash'
 import { DataGrid } from '@mui/x-data-grid'
+import { useSearchParams } from 'react-router-dom'
 
 const PAGE_SIZE = 15
+const DIRECTION = {
+  desc: '-',
+  asc: '+'
+}
+const DIRECTION_REVERSE = {
+  '-': 'desc',
+  '+': 'asc'
+}
 
 const Table = props => {
-  const [sortModel, setSortModel] = useState(props.initialSort)
+  const [searchParams, setSearchParams] = useSearchParams()
+  const [sortModel, setSortModel] = useState()
+  const [page, setPage] = useState()
+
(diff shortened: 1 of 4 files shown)


src/common/components/Table.js
```diff
@@ -1,18 +1,73 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
+import _ from 'lodash'
 import { DataGrid } from '@mui/x-data-grid'
 
 const PAGE_SIZE = 15
+const DIRECTION_BY_WORD = {
```
cacarrara, line 6: nitpick: I would recommend using `order` instead of `direction` here.

cacarrara (approved):

src/common/components/Table.js
josebui: @cacarrara I am going to change them to `SORT_DIRECTION_BY_WORD` and `SORT_DIRECTION_BY_SYMBOL` because it is the direction related to sort, I think this is clearer

PR #156, opened 2022-01-14
fix: describe Kobo as coming soon

### LINKED ISSUE #140: Describe KoboToolbox as coming soon
## Description
On the tools page, indicate that KoBoToolbox will be integrated into Terraso during Q1 2022 and users can log in to/use Kobo independently for now.

paulschreiber:
## Description
Adds text indicating Kobo will be available in the future. Move Kobo strings out of code an in to i18n JSON.

### Checklist
- [X] Strings are localized
- [X] Verified on mobile
- [X] Verified on desktop

### Related Issues
Fixes #140.

### Verification steps
Go to http://127.0.0.1:3000/tools/

### DIFF
--- a/src/localization/locales/en-US.json
+++ b/src/localization/locales/en-US.json
@@ -104,6 +104,7 @@
       "list_description": "Helpful tools in Terraso",
       "is_for": "This tool is for",
       "requirements": "System requirements",
+      "avilability": "Availability",
       "actions_bookmark": "Bookmark",
       "actions_recommend": "Recommend this to my team",
       "home_card_title": "Access Data Collection Tool in Terraso",
@@ -128,7 +129,8 @@
         ],
         "requirements": [
           "Modern web browser (Chrome, Edge, Firefox, Safari)"
-        ]
+        ],
+        "availability": "KoBoToolbox will be integrated into Terraso in 2022. Use Kobo independently for now."
       }
     },
     "localization": {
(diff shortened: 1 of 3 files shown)


src/localization/locales/en-US.json
```diff
@@ -112,6 +113,26 @@
       "home_card_kobo_title": "KoBoToolbox",
       "home_card_kobo_link": "Learn more about KoBoToolbox features"
     },
+    "tools": {
+      "kobo": {
+        "title": "KoBoToolbox",
+        "url": "https://www.kobotoolbox.org/",
+        "img": {
```
cacarrara, line 120: question: why image attributes need to be on the file with data to be translated?

josebui: thought: I am also not sure about this, but potentially localization is also related to how content displays, because on different locales depending on the content the image can use more or less space, not having another places to handle something like this related to each locale, I think this might be ok for now. My only concern is that translators might change this values breaking the application, but maybe we are small enough that this will not be an issue

paulschreiber: My thinking is the same as Jose's:
* The image in a different language could be a different size
* There isn't another good place to store the data; putting it in two places will make future updating more difficult.

cacarrara (approved):

josebui (approved):

PR #162, opened 2022-01-14
fix: align confirm and cancel buttons

paulschreiber:
## Description
Ensure the confirm button is rightmost.

### Before
image

### After
image


josebui:
@paulschreiber
Question: From where did this requirement came from? I see the buttons in Figma at the left side
Chore: Please add the screenshot of how this looks

paulschreiber:
@josebui in the [material UI spec](https://mui.com/components/dialogs/), the  confirm action is on the right.  I found [another exampler](https://dev.to/uguremirmustafa/material-ui-reusable-confirmation-dialog-in-react-2jnl) with MUI. You can also see this with macOS print dialogs:
image

josebui:
@paulschreiber that looks good to me, as long as Hana is also aligned with this that is ok for me

josebui (approved):

PR #168, opened 2022-01-19
feat: permissions check to display components

### LINKED ISSUE #167: Show / hide action components based on User permissions
## Description

Some actions are only allowed to specific users. For example: Group edit information is allowed only to Group managers. Soon, if a non-managers attempt to save a Group change, they will get an error message from the API.

To avoid compromising user experience, it's necessary to implement on web-client rules to show / hide action components based on what the current User can or cannot do. The known permission rules are the following:

- Only Group managers can edit Group information
- Only Group managers can remove a member from a Group
- Only Group managers can delete a Group

josebui:
## Description

- Added react `usePermission` hook and `<Restricted />` component for Permissions validation.

### Checklist
- [x] New tests added
- [x] Strings are localized
- [x] Verified on mobile
- [x] Verified on desktop

### Related Issues
Fixes #167 

### Verification steps

- Check the Restricted.test.js file


src/group/groupFragments.js
```diff
@@ -27,3 +27,15 @@ export const groupMembers = `
     }
   }
 `
+
+export const accountMembership = `
+  fragment accountMembership on GroupNode { 
+    accountMembership: memberships(user_Email_In: [$accountEmail]) {
```
cacarrara, line 33: Question: can't this information be fetched on the `groupMembers` fragment? :thinking:

src/group/groupFragments.js
```diff
@@ -27,3 +27,15 @@ export const groupMembers = `
     }
   }
 `
+
+export const accountMembership = `
```
cacarrara, line 31: Thought: `account` is a concept we don't have yet. Should we keep using `User` only? Do you see any problem using two words (user, account) to refer to the same thing (user)?

src/group/groupFragments.js
josebui: @cacarrara Is this question related to if this should be the correct place for this fragment? If that is it, the answer is yes since this is a fragment that applies over the `GroupNode` if that your question can you add more context to it?

src/group/groupFragments.js
josebui: @cacarrara I do think there is an account notion (at least from the web client side). So, I see 3 concepts referencing people from the web client:
Account: Anything related to the current logged in user, or the user performing the actions
Members: People related to a group (this can be inderectly members of the landscape too, through the default group)
Users: I think this will apply to user profile probabily when we get the requirement to access other people profiles

So, at least right now I don't see a confusion, when `account` is used it is linked to the current logged in user, and things mentioning users are refercing something to users in general, let me know where are you seeing this confusion

src/group/groupFragments.js
cacarrara: It was more about if we could get the member role on `groupMembers` and avoid heaving 2 fragments that get the same thing (memberships).

josebui: @cacarrara Ohh I see, let me check that

josebui: @cacarrara I checked. So, I think we should keep this as a separate fragment for some reasons:
- If for each group we fetch we always get all memberships then we will have to loop over them to match it with the current logged in user, adding extra work on the client
- When we add pagination, filtering for members, or just when we use the totalCount instead of getting the complete list of members we can't be sure that the membership for the current logged in user will be in that list
- Not every time we fetch the group data we need to know the role of the logged in user

So, this is why I think keeping a separate fragment just to get the information of the logged user account membership in the group is useful. Let me know if that makes sense

cacarrara (approved):

PR #180, opened 2022-01-25
feat: only show language picker if multiple languages are present

### LINKED ISSUE #130: Hide language switcher until multiple languages are enabled
## Description
Currently, the language switcher shows both English and Spanish, but only English is available.

- Hide the language switcher if only one language is enabled
- Indicate in a configuration file which languages are enabled

paulschreiber:
## Description
Only show language picker if at least two languages are present.

### Checklist
- [ ] New tests added
- [X] Verified on mobile
- [X] Verified on desktop

### Related Issues
Fixes #130.

### DIFF
--- a/src/localization/components/LocalePicker.js
+++ b/src/localization/components/LocalePicker.js
@@ -36,6 +36,11 @@ const LocalePicker = () => {
       ? `localization.locale_${locale}_small`
       : `localization.locale_${locale}`;
 
+  // Only show language picker if at least two languages are available.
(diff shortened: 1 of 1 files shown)


paulschreiber:
@josebui (1) Is this sufficient? Or do we want to check the contents of the JSON? (2) Should we add a test for this?

josebui:
@paulschreiber 
(1) I think this should be enough
(2) Knowing that locales are coming from a constant and we have now 2, I think test are not really relevant, as long as the picker is showing is tested with the existing tests I think we are good

PR #210, opened 2022-02-07
feat: Change title for each page, read content of new page

### LINKED ISSUE #200: Accessibility: Update <title> for each page
## Description
Each page should update the <title> tag value with the current page title

### LINKED ISSUE #208: Accessibility: Let users know when page has change
## Description
Right now when the user navigate to different pages there is no way to know something has changed. This is an opportunity tu use aria-live

josebui:
## Description

- Update page title on change route
- Read content when route changes

### Checklist
- [ ] New tests added
- [x] Strings are localized
- [x] Verified on mobile
- [x] Verified on desktop

### Related Issues
Fixes #200
Fixes #208


paulschreiber (changes requested):
issue: 
The `<title>` tag shouldn't match the `<h1>`.

Here are some titles we can use:

* home page should be `Terraso | Home`
* groups should be `Terraso | Group | Blah` (instead of `Terraso | Blah`). 
* landscapes should be `Terraso | Landscape | Blah` (instead of `Terraso | Blah`). 
* `Terraso | ACUS Penninsula Galera-San Francisco Members` should be `Terraso | Group | ACUS Penninsula Galera-San Francisco | Members`
* `Terraso | Manage "sample" Group Information` should be `Terraso | Group | Sample | Manage

For groups with long names, like `ACUS Peninsula Galera-San Francisco`, we should truncate them in the title … not sure how, though. @Hanal-code What do you recommend?

paulschreiber:
Terraso | Home 
Terraso | Groups
Terraso | Group | Blah | Edit
Terraso | Create Group
Terraso | Group | Blah | Members
Terraso | Group | Blah
Terraso | Landscapes
Terraso | Landscape | Blah | Members
Terraso | Landscape | Blah
Terraso | Tools
Terraso | Log in
Terraso | Account

Hanal-code:
I'd suggest to put the current page title first instead of 'Terraso' to provide descriptive titles based on [W3C document](https://www.w3.org/WAI/WCAG21/Techniques/general/G88) 
When many tabs open, users will see 'Terraso | ...' only which already indicated by favicon. 

The current-page-first approach will make the page title;
Home | Terraso 
Groups  | Terraso 
Edit: Blah | Terraso
Create Group | Terraso
Members: Blah  | Terraso
Blah | Terraso
Landscapes | Terraso
Tools | Terraso
Log in | Terraso
Account | Terraso

I don't recommend to use the page title like a breadcrumb which will be added to the app later on.

paulschreiber:
I like @Hanal-code 's suggestions! Those are an improvement. One change I would make is for the home page to lead with `Terraso` — perhaps just `Terraso` or `Terraso Home`?

josebui:
Thanks @Hanal-code @paulschreiber . Hana let me know what do you think about Paul's suggestion for the home title

Hanal-code:
@paulschreiber  Can you elaborate your rationale to deviate the pattern for the Home page?

paulschreiber:
@Hanal-code starting with `Terraso` allows a good title for typeahead completion and bookmarks.

src/common/document.js
```diff
@@ -1,12 +1,19 @@
...
-    : `${title} | ${t('common.terraso_projectName')}`;
+  const titleParts = [
+    fetching ? null : title,
+    omitSuffix ? null : t('common.terraso_projectName'),
+  ];
+
+  const fullTitle = _.flow(
+    _.filter(part => !_.isEmpty(part)),
+    _.join(' | ')
+  )(titleParts);
```
paulschreiber, line 16: suggestion: simplify this as `_,compact(titleParts).join(' | ')`.

paulschreiber (changes requested):

src/common/document.js
josebui: Thanks, didn't know about that compact function

paulschreiber (changes requested):
issue: when changing pages, the title changes to `Terraso` for a brief period. Please eliminate that extra change.

josebui:
@paulschreiber That happens because the app is loading data, what should be the title while the app doesn't have the information to display the correct title?

src/common/document.js
paulschreiber: Any time you are filtering on "not empty" in a verbose way, it's a hint to look for something briefer. :)

In regular JavaScript, you can also do `foo.filter(bar => bar)` and `foo.filter(Number)`.

paulschreiber (approved):

PR #215, opened 2022-02-10
chore: adds Leaflet dependency

cacarrara:
This change adds Leaflet as a project dependency. As a good consequence,
we can now avoid loading Leaflet CSS from CDN and load from project
dependency directly.


src/gis/components/Map.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import { MapContainer, TileLayer } from 'react-leaflet';
 
+import './Map.css';
```
josebui, line 4: Suggestion:
Just a suggestion, for me is better to keep these imports always explicit and not relative
```suggestion
import 'gis/components/Map.css';
```

josebui (approved):
Looks good, just one suggestion

Issue #192, opened 2022-02-01
Display landscape polygon on map


paulschreiber:
## Description
On landscape profile pages, render the polygon showing the landscape boundary. (Previously, we were using a string to find the location.)

PR #227, opened 2022-02-14
feat: add Landscape area polygon to the map 

cacarrara:
This change updates the Map component to support it receiving a GeoJSON
objects. This GeoJSON is used to add a new map polygon layer.

The new `geojson` property is already being used to the Landscape's
`areaPolygon` attribute. When a `areaPolygon` is used, the bounding box
of this polygon is used. Otherwise, the Landscape's position bounding
box continue being used.

Even before this change, the `LandscapeView` had a limitation to don't
properly handle the situation where a Landscape has no location. This
limitation isn't addressed in this change.

Fix: #192 

image

### DIFF
--- a/src/landscape/components/LandscapeView.js
+++ b/src/landscape/components/LandscapeView.js
@@ -14,6 +14,7 @@ import {
   Box,
 } from '@mui/material';
 import PublicIcon from '@mui/icons-material/Public';
+import bbox from '@turf/bbox';
 
 import { fetchLandscapeView } from 'landscape/landscapeSlice';
 import { withProps } from 'react-hoc';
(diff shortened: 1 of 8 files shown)


paulschreiber (approved):

src/landscape/components/LandscapeView.js
```diff
@@ -68,16 +69,24 @@ const LandscapeCard = ({ landscape }) => {
   );
 };
 
-const LandscapeMap = ({ position }) => {
+const LandscapeMap = ({ landscape }) => {
   const { t } = useTranslation();
-  const bounds = position && [
-    [position.boundingbox[0], position.boundingbox[2]],
-    [position.boundingbox[1], position.boundingbox[3]],
+  const { areaPolygon, position } = landscape;
```
josebui, line 74: suggestion: I think this logic to calculate the bounds can be extracted to a separate file, maybe to `src/landscape/polygonUtils.js` with this also unit tests could be added to validate how the areaPolyong or poistion is transform to the bounds value

josebui (approved):
Nice, we have polygons now

PR #249, opened 2022-02-24
feat: Not found page

### LINKED ISSUE #183: Error page (404 and 500 errors)
## Description
Create error page templates/pages for 404 and 500 errors.
Non Engr tasks: https://trello.com/c/efmFDKjR/40-list-global-error-state-and-draft-helpful-error-messages

MVP:
- [x]  simple 404 error page design
- [ ] implementation

josebui:
## Description

- Added not found page

### Checklist
- [ ] New tests added
- [x] Strings are localized
- [x] Verified on mobile
- [x] Verified on desktop

### Related Issues
Fixes #183 

### Verification steps

Screenshots:

download (47)
download (46)


src/layout/NotFound.js
```diff
@@ -0,0 +1,63 @@
...
+              marginTop: 1,
+              fontSize: '16px',
+              textTransform: 'uppercase',
+            })}
+          >
+            {t('common.not_found_error_code')}
+          </Typography>
+          <Typography sx={{ marginTop: 4, fontSize: '16px' }}>
+            <Trans i18nKey="common.not_found_help_message">
+              Prefix <Link href={HELP_URL}>link</Link>.
```
cacarrara, line 40: Doubt: when this `Prefix ...` message will be used?

cacarrara (approved):
Overall it LGTM :+1:

src/navigation/Navigation.js
```diff
@@ -13,15 +13,15 @@ const PAGES = {
...
   },
   '/groups': {
     label: 'navigation.groups',
-    match: path => path.startsWith('/groups'),
+    match: path => path === '/groups' || path.startsWith('/groups/'),
   },
   '/tools': {
     label: 'navigation.tools',
-    match: path => path.startsWith('/tools'),
+    match: path => path === '/tools' || path.startsWith('/tools/'),
```
paulschreiber, line 24: question: why doesn't `startsWith('/tools'`)` do what we want?

paulschreiber (changes requested):
issue: let's move the body context so it is left-aligned, like the other pages are.

image

src/layout/NotFound.js
josebui: @cacarrara It is just to identify the parts for the translation, it is used by i18next to identify the parts that will be replaced on each locale

src/navigation/Navigation.js
josebui: @paulschreiber `startsWith('/tools')` picks anything that starts with `/tools` but that is not exaclty what we need. The user is under or in `tools` if the route is exactly `/tools` or under `/tools/` but for example a different thing that start with `tools` like `tools123` should not be identified as `tools`, does that make sense?

src/layout/NotFound.js
josebui: This one has 3 parts `prefix` `link` `period`

src/navigation/Navigation.js
paulschreiber: What about something like`startsWith('/tools(/.+)?')`?

josebui: sure, a regular expression might be able to handle this too, I will test that

paulschreiber (approved):

PR #300, opened 2022-04-13
fix: Disable link on logo when on home page

### LINKED ISSUE #294: Disable link on logo when on home page
## Description
Currently, the Terraso logo links to the home page. When we are already on the home page, the logo should not be clickable.

paulschreiber:
## Description
Disables the link on the logo when on home page (don't self-link).

### Checklist
- [X] Verified on mobile
- [X] Verified on desktop

### Related Issues
Fixes #294.

### DIFF
--- a/src/common/components/ConditionalLink.js
+++ b/src/common/components/ConditionalLink.js
@@ -1,0 +1,11 @@
+import { Link } from 'react-router-dom';
+
+const ConditionalLink = ({ children, to, condition }) => {
+  if (!!condition && to) {
+    return <Link to={to}>{children}</Link>;
+  } else {
+    return <>{children}</>;
+  }
+};
+
+export default ConditionalLink;
--- a/src/common/components/AppBar.js
+++ b/src/common/components/AppBar.js
@@ -4,6 +4,7 @@ import { AppBar, Toolbar, Button, Box } from '@mui/material';
 import { Link } from 'react-router-dom';
 import { useSelector, useDispatch } from 'react-redux';
 import { useTranslation } from 'react-i18next';
+import { useLocation } from 'react-router-dom';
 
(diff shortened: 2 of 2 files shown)


src/common/components/ConditionalLink.js
```diff
@@ -0,0 +1,11 @@
+import { Link } from 'react-router-dom';
+
+const ConditionalLink = ({ children, to, condition }) => {
+  if (!!condition && to) {
+    return <Link to={to}>{children}</Link>;
+  } else {
+    return <>{children}</>;
+  }
```
josebui, line 8: Suggestion: This feels easier to read for me
```suggestion
  return (!!condition && to)
    ? <Link to={to}>{children}</Link>
    : <>{children}</>;
```

src/common/components/ConditionalLink.js
```diff
@@ -0,0 +1,11 @@
+import { Link } from 'react-router-dom';
+
+const ConditionalLink = ({ children, to, condition }) => {
+  if (!!condition && to) {
+    return <Link to={to}>{children}</Link>;
+  } else {
+    return <>{children}</>;
```
josebui, line 7: Question: Should this be wrapped by `Typography`? I think it will be just text directly in the markup without it.

paulschreiber: Not necessarily. In this case, I have an image, which doesn't need a wrapper.

src/common/components/ConditionalLink.js
paulschreiber: The [Stack Overflow answer](https://stackoverflow.com/a/56278522) had a ternary, but I found it to be overly dense/harder to follow. Spreading it out is more readable and will make debugging easier.

josebui (approved):

PR #310, opened 2022-04-19
feature: add search to map component

### LINKED ISSUE #299: Landscape creation: add leaflet map with pin drop tool
* when loading page for initial creattion, center map on country
* when saving, save both lat/lng of pin and zoom level

image

paulschreiber:
## Description
Add a search box to map component. When a location is found, a draggable pin is inserted.

### Checklist
- [ ] New tests added
- [X] Verified on mobile
- [X] Verified on desktop

### Related Issues
Fixes #299.

### DIFF
--- a/src/gis/components/Map.js
+++ b/src/gis/components/Map.js
@@ -1,8 +1,81 @@
-import React, { useEffect } from 'react';
+import React, { useState, useEffect } from 'react';
 import { GeoJSON, MapContainer, TileLayer, useMap } from 'react-leaflet';
 import { v4 as uuidv4 } from 'uuid';
+import { GeoSearchControl, OpenStreetMapProvider } from 'leaflet-geosearch';
+import L from 'leaflet';
+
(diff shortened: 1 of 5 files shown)


src/gis/components/Map.js
```diff
@@ -1,9 +1,79 @@
...
 import { v4 as uuidv4 } from 'uuid';
+import { GeoSearchControl, OpenStreetMapProvider } from 'leaflet-geosearch';
+import L from 'leaflet';
+
+import 'leaflet/dist/leaflet.css';
+import 'leaflet-geosearch/dist/geosearch.css';
 
 import 'gis/components/Map.css';
 
+delete L.Icon.Default.prototype._getIconUrl;
```
josebui, line 12: Question: Why is this needed?

josebui (approved):
Looks good, just one question

src/gis/components/Map.js
paulschreiber: If we don't include it, the marker icon is broken.

PR #320, opened 2022-04-26
fix: disable rollbar when token is not present

### LINKED ISSUE #318: Disable rollbar in local development environment
## Description
Disable rollbar in the local development environment. If a rollbar token is not present, don't call the API.

paulschreiber:
## Description
Disable rollbar when token is not present.

### Related Issues
Fixes #318.


josebui (approved):
Nice, that is a good small change

PR #321, opened 2022-04-26
feat: Landscape creation

### LINKED ISSUE #296: Add create landscape button and text to landscapes index page
image

### LINKED ISSUE #297: Add Landscape creation screen
image

* Decide if country list should be compiled into client or fetched from server

### LINKED ISSUE #298: After landscape creation, show landscape page with confirmation message

### LINKED ISSUE #299: Landscape creation: add leaflet map with pin drop tool

josebui:
## Description

- Landscape creation

### Checklist
- [x] New tests added
- [x] English strings reviewed and copyedited
- [x] Strings are localized
- [x] Verified on mobile
- [x] Verified on desktop

### Related Issues
Fixes #299 
Fixes #298 
Fixes #296 
Fixed #297


Hanal-code:
**UX verification feedback:** 

1. In Step2. the map (on page load) was not oriented to the country user selected in step1. 
2. The county list (in step 1) opens too wide & big; the appearance is not a drop-down list. 
3. The country or Region list (in step 1) doesn't seem to cover names of region. Expect the label and the options to match. 
4. The pin tool in Drop a pin is not available. It requires the user (1) activate the search field (2) enter keyword (3) Drag the pin to the right location. These interaction doesn't match to the design, so I assume there were some tradeoffs engineers made during the implementation. I would like to hear that and be on the same page. 
5. Improve the Website input field or the error message; first, the error message is misleading as the error is not about the validity of the web address. The error is about the format (or complete URL). For example, 'www.techmatters.org' erred because of missing 'http://' or 'https://'  Can we change it to accept the URL without 'http...'? OR the error message should give the format example in the message. 
6. The search field in the Drop a pin screen is expected to be visible. The current implementation requires the user to activate to see the input field. 
7. Disabled state of CTA button only appears in the step 2. Do not block the user to proceed toward task completion with optional user entry. 
8. The zoom level in the Drop pin screen didn't match to the zoom level of the landscape map. 
9. Roads were highlighted in red including exit numbers. Remove unnecessary info from the map.  
10. Enlarge the map to 100% of the main (2 columns instead of 1 column)

paulschreiber:
>The country or Region list (in step 1) doesn't seem to cover names of region. Expect the label and the options to match.

The wording is "country or region" to avoid disputes over what is or is not a country, i.e. Taiwan. It is not intended to display states/provinces or other second-level administrative districts.

>Improve the Website input field or the error message; first, the error message is misleading as the error is not about the validity of the web address. The error is about the format (or complete URL). For example, 'www.techmatters.org' erred because of missing 'http://' or 'https://' Can we change it to accept the URL without 'http...'? OR the error message should give the format example in the message.

We should accept URLs without protocols and fix them on save. We did this for groups already.

Hanal-code:
Follow-up meeting note: https://docs.google.com/document/d/1Ukgmxh8NRjUj_pmBXzEasdW0IkiWch1VHDqnxUkpzyc/edit#

Hanal-code:
UX verification feedback:
- [Drop pin screen] Activation of the pin - initial location of the pin display should be outside of the button. Cancel Pin, the pin tool tips are overlaid.
image

- [Drop pin screen] Pin icon needs to be larger. Hard to discern what it is. 
- [Drop pin screen] move the search field to top left and add the pin button nearby. 
Refer to the mockup: 
image

- [Drop pin screen] keyboard navigation focus order is confusing. The current order is 1)Zoom controls, 2)Pin, 3)Leftlet links 4)Search. 
Expected order: 1) search field  2) Pin  3)Zoom controls 

- Closing the location search suggestion requires clicking 'x' - dismiss it when hit 'Enter' keystroke. 
image

- Is it possible to drop a pin on page load at the center of the map? 

- [Mobile] On the form page load, the form page shows the bottom the page as if it's stuck. Expected to see the top of the page with H1. 
- [Mobile] How to move the pin in mobile? Rendering the map tile is slow and the selecting the pin is not clear.

paulschreiber:
>Closing the location search suggestion requires clicking 'x' - dismiss it when hit 'Enter' keystroke.

Clicking on the name also hides the menu. Enter _should_ work, though.

src/gis/components/Map.js
```diff
@@ -19,19 +109,104 @@ const MapPolygon = props => {
...
+      whenReady={setMap}
+      zoom={3}
+      center={[0, 0]}
+      {..._.omit(['center', 'zoom'], props)}
     >
       <TileLayer
-        attribution='&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
-        url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
+        attribution='Data &copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors Tiles &copy; HOT'
+        url="https://{s}.tile.openstreetmap.fr/hot/{z}/{x}/{y}.png"
```
josebui, line 201: Question: @paulschreiber why are we using the french url here?

paulschreiber: This is the "humanitarian" tileset (HOTOSM). Hana preferred this art to the standard art.

paulschreiber (approved):

PR #326, opened 2022-04-28
chore: sort imports

paulschreiber:
## Description
Sorts the imports at the beginning oof each file. Adds a prettier plugin and configuration to handle the sorting.

Order is:
* tests/utils
* react
* @mui
* lodash
* third-party modules
* local modules

Tests will fail if `tests/utils` is imported too late, so all of those imports are moved to the top.

### DIFF
--- a/src/account/components/AccountProfile.test.js
+++ b/src/account/components/AccountProfile.test.js
@@ -1,10 +1,12 @@
-import AccountProfile from 'account/components/AccountProfile';
-import _ from 'lodash/fp';
+import { fireEvent, render, screen, within } from 'tests/utils';
+
 import React from 'react';
 import { act } from 'react-dom/test-utils';
-import * as terrasoApi from 'terrasoBackend/api';
 
-import { fireEvent, render, screen, within } from 'tests/utils';
+import _ from 'lodash/fp';
+
+import AccountProfile from 'account/components/AccountProfile';
+import * as terrasoApi from 'terrasoBackend/api';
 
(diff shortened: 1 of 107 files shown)


src/account/components/AccountProfile.test.js
```diff
@@ -1,11 +1,11 @@
-import React from 'react';
+import AccountProfile from 'account/components/AccountProfile';
```
josebui, line 1: thought: This doesn't seem better to me, this is mixing Visual Components, with non visual dependencies, and owned dependencies with thrid party dependencies.

question: What are the advantages you see on this approach?

paulschreiber: The idea is to be consistent across files and, within a file, across functions within an import. We already do this in our Python code.

If the default order isn't good, we can customize it.

src/common/components/AppBar.js
```diff
@@ -1,19 +1,22 @@
...
+import { Link } from 'react-router-dom';
 import { useLocation } from 'react-router-dom';
 
-import LocalePicker from 'localization/components/LocalePicker';
-import { signOut } from 'account/accountSlice';
-import theme from 'theme';
+import { AppBar, Box, Button, Toolbar } from '@mui/material';
+import useMediaQuery from '@mui/material/useMediaQuery';
 
-import logo from 'assets/logo.svg';
```
josebui, line 13: Suggestion: I think assets should be closer to theme and css at the end

josebui (approved):

PR #331, opened 2022-05-06
feat: Upload component for shared data files

josebui:
## Description

- Upload page

### Checklist
- [ ] New tests added
- [ ] English strings reviewed and copyedited
- [ ] Strings are localized
- [ ] Verified on mobile
- [ ] Verified on desktop
- [ ] Verified accessibility ([Accessibility - Web applications](https://docs.google.com/document/d/1VwPDyLqw7r_iISMgDZr1FdAmAqqOIIVIxQqsYA-HaHE/edit?usp=sharing))


src/common/components/DropZone.js
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+
+import { useDropzone } from 'react-dropzone';
+import { useTranslation } from 'react-i18next';
+
+import { Alert, Paper, Stack, Typography } from '@mui/material';
+
+import theme from 'theme';
+
+const getFormatedSize = bytes => {
```
paulschreiber, line 10: issue: use the filesize library here, as I do in the FileCard component.

paulschreiber (changes requested):

src/common/components/DropZone.js
```diff
@@ -0,0 +1,122 @@
...
+            })}
+          >
+            {t('common.drop_zone_select_file')}
+          </Paper>
+          <Typography
+            variant="caption"
+            sx={{ fontWeight: 'bold', paddingTop: 1 }}
+          >
+            {t('common.drop_zone_format', {
+              extensions: fileExtensions.map(ext => `*.${ext}`).join(', '),
```
paulschreiber, line 103: question: should we call `map()` outside the component and place the value in a `const`?

src/config.js
```diff
@@ -16,6 +16,15 @@ export const AXE_TEST_TIMEOUT = process.env.REACT_AXE_TEST_TIMEOUT || 20000;
 
 export const GEOJSON_MAX_SIZE = process.env.REACT_GEOJSON_MAX_SIZE || 1000000;
 
+export const SHARED_DATA_MAX_SIZE =
+  process.env.REACT_SHARED_DATA_MAX_SIZE || 10000000;
```
paulschreiber, line 20: suggestion: for these large values, add a comment with the human-readable size, i.e `// 10 MB`

src/localization/locales/en-US.json
```diff
@@ -244,15 +244,28 @@
         "form_boundary_geojson_title": "Upload your landscape’s boundary",
         "form_location_placeholder": "Landscape location",
         "form_location_select": "Select location",
-        "form_name_placeholder": "Your landscape’s official name"
+        "form_name_placeholder": "Your landscape’s official name",
+        "boundaries_file_no_accepted": "No accepted files",
```
paulschreiber, line 248: issue: Need to better explain what the error is.

src/localization/locales/en-US.json
```diff
@@ -244,15 +244,28 @@
...
     },
     "shared_files": {
         "title": "Shared files",
         "description": "Only the members of your group can download the files. Read more about sharing files.",
         "upload_button": "Upload Files",
         "delete_confirm_title": "Delete “{{name}}”?",
         "delete_confirm_message": "Are you sure you wish to delete the file “{{name}}”?",
-        "delete_confirm_button": "Delete File"
+        "delete_confirm_button": "Delete File",
+        "upload_file_success": "[TODO] File upload succesfuly",
```
paulschreiber, line 258: question: is this the inline success message (above the file name)?

src/localization/locales/en-US.json
```diff
@@ -244,15 +244,28 @@
...
+        "upload_filename_label": "File Name",
```
paulschreiber, line 259:
```suggestion
        "upload_filename_label": "Name",
```

src/localization/locales/en-US.json
```diff
@@ -244,15 +244,28 @@
...
+        "upload_description_label": "File Description",
```
paulschreiber, line 260:
```suggestion
        "upload_description_label": "Description",
```

src/localization/locales/en-US.json
```diff
@@ -244,15 +244,28 @@
...
+        "upload_description_placeholder": "[TODO] e.g. Soil health survey data",
+        "upload_no_files": "No files added",
+        "upload_rejected_too-many-files": "[TODO] Too many files. Max number of files: {{maxFiles}}",
+        "upload_rejected_file-invalid-type": "[TODO] Invalid file type. Accepted extensions: {{fileExtensions}}. Files: {{rejectedFiles}}",
+        "upload_rejected_file-too-large": "[TODO] File too large. Max size: {{maxSize}}MB. Files: {{rejectedFiles}}",
+        "upload_description": "Add one or more data files (xls, xlsx, csv files only) and descriptions",
```
paulschreiber, line 266:
```suggestion
        "upload_description": "Add one or more data files (xls, xlsx, csv) and descriptions",
```

paulschreiber (changes requested):

src/localization/locales/en-US.json
josebui: @paulschreiber This is the message you added in the Boundaries dropzone when there are no accepted files, I just moved this to be translated, see: https://github.com/techmatters/terraso-web-client/pull/331/files#diff-378e736e2329b798aa6afb45a118911249c370f297dd595d9998bea24b1c1092L75 . What should be the text here?

src/localization/locales/en-US.json
josebui: @paulschreiber yes, this will be the succes message on top of each file

src/common/components/DropZone.js
josebui: @paulschreiber Can you add more context on why you think that will be better? Me reasoning here was:
This is only specifically needed for `common.drop_zone_format`, so there is no value on having the joined extensions string far from the localized string parameters, I think having it close can potentialy let us improve this (I am not sure if I like the join here completely, but that can be a future improvement), and the scope of action of this is really clear.

PR #340, opened 2022-05-13
fix: reorder map controls per spec

paulschreiber:
## Description
Makes search field top left and pin to the right of that.

image

### Checklist
- [X] Verified on mobile
- [X] Verified on desktop

### DIFF
--- a/src/gis/components/Map.js
+++ b/src/gis/components/Map.js
@@ -59,7 +59,7 @@ const LeafletDraw = props => {
     });
 
     return () => map.removeControl(drawControl);
-  }, [map, setPinLocation]);
+  }, [map, setPinLocation, isSmall]);
   return null;
 };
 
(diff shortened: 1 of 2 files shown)


src/gis/components/Map.js
```diff
@@ -206,7 +214,9 @@ const Map = props => {
         enableSearch={props.enableSearch}
         enableDraw={props.enableDraw}
       />
-
+      {'undefined' !== typeof props.enableSearch && (
```
josebui, line 217: Question: Do we need this? null and undefined are falsy so I think that should be enough for this

paulschreiber: I had inverted logic originally and "fixed" it this way. Thanks for flagging.

paulschreiber, line 217:
```suggestion
      {props.enableSearch && (
```

PR #417, opened 2022-06-13
fix: make footer same width as the rest of the page

paulschreiber:
## Description
On desktop, make the footer same width as the rest of the page

### Before
image

### After
image


josebui:
@paulschreiber can you add an screenshot of how this looks?

josebui (approved):

PR #419, opened 2022-06-15
chore: React 18 upgrade

### LINKED ISSUE #335: Update to React 18
## Description
Update the web cline to use React 18.

josebui:
## Description

- React 18 Upgrade

### Checklist
- [x] New tests added
- [x] English strings reviewed and copyedited
- [x] Strings are localized
- [x] Verified on mobile
- [x] Verified on desktop

### Related Issues
Fixes #335


josebui:
It seems that Mui is not explecitely ready for React 18, but as far as they know it is compatible https://github.com/mui/material-ui/issues/29844

paulschreiber (approved):

josebui:
Merging to main still blocked by `@emotion` dependencies, they are still not ready to work with React 18

src/state/store.js
```diff
@@ -8,8 +9,21 @@ import userHomeReducer from 'home/homeSlice';
 import landscapeReducer from 'landscape/landscapeSlice';
 import sharedDataReducer from 'sharedData/sharedDataSlice';
 
+const handleAbortMiddleware = store => next => action => {
```
paulschreiber, line 12: question: what is this used for?

src/state/utils.js
```diff
@@ -73,11 +77,23 @@ export const createAsyncThunk = (
...
+      if (dispatchErrorMessage && !isAborted) {
         parsedErrors.forEach(message => dispatch(addMessage(message)));
       }
 
       return rejectWithValue({ error, parsedErrors });
     }
   });
 };
+
+export const useFetchData = dataFetchCallback => {
```
paulschreiber, line 89: question: every time this is called, it wraps useCallback. i.e.
```
 useFetchData(useCallback(() => foo
```

Is there a way to write it so you can avoid the extra layer, i.e.
```
 useFetchData(() => foo
```
?

paulschreiber (approved):

src/state/utils.js
josebui: @paulschreiber Yes, that is required, react hooks have to be used under the React components directly, this si a good practice and important to validate the ordering of their execution and theri dependencies, so still this is needed not sure if this will be improved in the future.

src/state/store.js
josebui: @paulschreiber This is used to indetify a new type of request execution result when a request is aborted, so that we can handle that and avoid showing the user the result of request no longer relevant

paulschreiber: Can you tell me more? What changed about React 18 to require this?

josebui: @paulschreiber This is also related to the double mounting of components that created the need for the `useFetchData` hook, so what react 18 is doing in development mode is mount and umount the component, that is when (for some route components) the calls to the API are triggered more than one time, and in the middle of that mount -> unmount -> mount there is an aborted request to the api that needs to be handled

PR #483, opened 2022-08-10
fix: improve color contrast for placeholder text, hover state, map

### LINKED ISSUE #482: Color contrast below required ratio
## Description
In several places, the color contrast ratio does not meet the WCAG guidelines.

### Insufficient Color Contrast for Landscape Boundaries
The boundaries of the landscape are indicated on the map using blue lines (#3388FF) that do not meet the minimum color contrast requirements for some parts of the map. For example, when displayed on blue water bodies (#90CCCB), the resulting contrast ratio is 1.9:1 which fails the minimum requirement of 3:1 for graphical objects.

Use a darker blue color to mark the landscape location to ensure a contrast ratio of at least 3:1 with all terrains. One suggestion is using #0055CC which will result in sufficient contrast ratio with all parts of the map.

### Landscape Details Form: Insufficient Color Contrast for Placeholder Text
The placeholder text for form fields is using a light gray color (#AAAAAA) on a white (#FFFFFF) background. This results in a color contrast ratio of 2.32:1, below the 4.5:1 minimum requirement for normal sized text. The same gray color is used for the default value "Landscape Location" for "Country or Region" dropdown.

A darker gray color must be used to ensure that a contrast ratio of at least 4.5:1 is attained. For instance, #757575 would achieve a contrast ratio of 4.61:1 on the white background.

### Landscape Details Form: Insufficient Color Contrast for Input Boundaries
The form input fields have a gray boundary (#C4C4C4) on a white background (#FFFFFF) resulting in a contrast ratio of 1.74:1, below the 3:1 minimum requirement for boundaries indicating the hit area of input fields. Note that some fields have a placeholder text which is a sufficient indicator of a field hit area, but other fields e.g., "Description" has no placeholder text and relies solely on the gray boundary.

Use a darker gray color for the boundaries. One suggestion is using #949494 to achieve a contrast ratio of 3.03:1 against the white background. 

### Member: Insufficient Color Contrast for Buttons on Hover 
When any "Connect" button in the "Actions" column of the landscapes grid is activated, the button is replaced with a "Member" button. On hover, the button has black text (#000000) on a dark blue background (#21586D) resulting in a contrast ratio of 2.68:1 which fails the minimum contrast requirement of 4.5:1 for text.

The same styling is applied on the "Manager" button for a landscape that has been created by the user.

Use a lighter shade of blue for the background that will result in a contrast ratio of at least 4.5:1 with the black text. One suggestion is using #2F7E9D that has a contrast ratio of 4.59:1.

### Table: Insufficient Color Contrast for Links on Hover
Links in the landscapes grid have a blue color (#307F9C). On mouse hover, the grid row changes its color from white to light gray (#F5F5F5) resulting in a contrast ratio of 4.1:1 with the link text which fails the minimum contrast requirement of 4.5:1 for text.

The same link color is used throughout the website, so a better approach is keeping the background white and providing another hover effect such as an underline.

### Home: Insufficient Color Contrast for Links on Hover
On hover, the "Explore Landscapes" and "Explore Groups" links have a light blue background (#F7FAFB) that has a contrast ratio of 4.31:1 against the blue text (#307F9C) which fails the minimum contrast requirement of 4.5:1 for regular text.

The same issue is observed for the "Connect to Another Landscape" and "Join Another Group" links that are displayed when user has created a landscape or a group.

Use a different hover effect for the links e.g., making the text darker or adding an underline styling.

paulschreiber:
## Description
Improves the color contrast ratio does to meet the WCAG guidelines.

### Related Issues
Fixes #482.

### DIFF
--- a/src/common/components/Actions.js
+++ b/src/common/components/Actions.js
@@ -8,7 +8,14 @@ const Actions = ({ label, destination }) => {
       <Button
         component={Link}
         to={destination}
-        sx={{ width: '100%', textTransform: 'uppercase' }}
+        sx={{
+          width: '100%',
+          textTransform: 'uppercase',
+          '&:hover': {
+            backgroundColor: 'transparent',
+            textDecoration: 'underline',
+          },
+        }}
       >
         {label}
       </Button>
--- a/src/common/components/Actions.js
+++ b/src/common/components/Actions.js
@@ -1,0 +1,19 @@
+import React from 'react';
+
+import { Button, CardActions, Link } from '@mui/material';
+
+const Actions = ({ label, destination }) => {
+  return (
+    <CardActions>
+      <Button
+        component={Link}
+        to={destination}
+        sx={{ width: '100%', textTransform: 'uppercase' }}
+      >
+        {label}
+      </Button>
+    </CardActions>
+  );
+};
+
+export default Actions;
--- a/src/common/components/CardActionButton.js
+++ b/src/common/components/CardActionButton.js
@@ -2,7 +2,7 @@ import React from 'react';
 
 import { Button, CardActions, Link } from '@mui/material';
 
-const Actions = ({ label, destination }) => {
+const CardActionButton = ({ label, destination }) => {
   return (
     <CardActions>
       <Button
@@ -23,4 +23,4 @@ const Actions = ({ label, destination }) => {
   );
 };
 
-export default Actions;
+export default CardActionButton;
--- a/src/common/components/CardActionButton.js
+++ b/src/common/components/CardActionButton.js
@@ -1,13 +1,15 @@
 import React from 'react';
 
-import { Button, CardActions, Link } from '@mui/material';
+import { Link } from 'react-router-dom';
 
-const CardActionButton = ({ label, destination }) => {
+import { Button, CardActions } from '@mui/material';
+
+const CardActionButton = ({ label, to }) => {
   return (
     <CardActions>
       <Button
         component={Link}
-        to={destination}
+        to={to}
         sx={{
           width: '100%',
           textTransform: 'uppercase',
--- a/src/common/components/TableResponsive.js
+++ b/src/common/components/TableResponsive.js
@@ -23,8 +23,6 @@ import {
 import BaseTable from 'common/components/Table';
 import ResponsiveSwitch from 'layout/ResponsiveSwitch';
 
-import theme from 'theme';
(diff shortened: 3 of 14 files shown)


src/common/components/Actions.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+import { Button, CardActions, Link } from '@mui/material';
+
+const Actions = ({ label, destination }) => {
```
josebui, line 5: Issue: I don't think this component name is describing its intent. It seems to be a specific component to handle navigation links inside cards that only hold one action.

Suggestion: Make this just a customization of `Button` (Remove `CardActions` from this file) and improve the name related to the navigation intent

josebui (changes requested):

src/common/components/TableResponsive.js
```diff
@@ -259,8 +261,12 @@ const SearchBar = props => {
         sx={{ flexGrow: 2, width: { xs: '100%' } }}
         InputProps={{
           sx: {
+            borderColor: theme.palette.gray.dark2,
             marginBottom: 2,
             padding: 0,
+            '& ::placeholder': {
+              opacity: 0.65,
+            },
           },
```
josebui, line 270: Suggestion: You can get the `theme` from inside the sx property without importing it
```suggestion
          sx: theme => ({
            borderColor: theme.palette.gray.dark2,
            marginBottom: 2,
            padding: 0,
            '& ::placeholder': {
              opacity: 0.65,
            },
          }),
```

src/common/components/Actions.js
paulschreiber: How about `CardActionButton`? Is that better? If not, can you customize Button as you would like it?

src/common/components/CardActionButton.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+import { Button, CardActions, Link } from '@mui/material';
+
+const CardActionButton = ({ label, destination }) => {
```
josebui, line 5: Issue: This name is still not describing the intent of the component, I think `CardActionRouterLink` is better. It seems in here you are using the `Link` component from `@mui/material` and the expectation from the places where you replace this is to use `import { Link } from 'react-router-dom';`. Is this still working as expected?

josebui (changes requested):

josebui (approved):

PR #496, opened 2022-08-29
fix: make markup of share dialog and landscape more accessible

### LINKED ISSUE #495: Accessibility markup errors, part 1
## Description
Areas in which markup is not accessible:

### Incorrect Focus on Dialog Display
When the "Share" button is activated, focus is set on a parent container wrapping the dialog (`<div>` with class="MuiDialog-container MuiDialog-scrollPaper css-ekeie0"). This causes inconsistent or missing announcements across environments and some screen reader users will not be informed that a dialog was opened.

Manage the focus so that it is placed inside the modal overlay. Send focus to the `<button>` with class="MuiButtonBase-root MuiIconButton-root MuiIconButton-sizeMedium css-1c3vy6f" that is used to close the dialog.

### Missing Accessible Name for Close Button
The close ("X") button has no accessible name to convey its purpose to screen reader users.

Add an aria-label attribute to the button e.g., aria-label="Close", to provide a descriptive name for the button.

### Missing Programmatic Label for Text Field
The "Copy the link to share" field is not programmatically associated with its visible label. When screen reader users navigate to the field, they won't be given any indication of its purpose.

Use a `<label>` element instead of the existing `<p>`, along with a "for" attribute to associate the label with the field. Ensure that the "for" attribute on the `<label>` and the ID of the corresponding field match.

### Incorrect Heading Structure
The create landscape process has several steps displayed on sequential screens, each step heading is implemented as a level 1 heading, however, there is no main heading describing the main purpose of the page. Additionally, the steps progress bar is placed before the currently used level 1 heading and may be missed by screen reader users navigating through the page using headings.

Add a level 1 heading with the text "Create Landscape" as the first element in the main content and implement each step heading as a level 2 heading instead of the currently used level 1 heading.

### Missing List Structure for Buttons
The sharing buttons are implemented as individual buttons with no list mark-up to convey their structure. Screen reader users will not be informed that the buttons are grouped and how many items are in the collection.

Encompass the buttons in an unordered list `<ul>` with `<li>` wrapping each button. Use the aria-labelledby attribute to reference the ID of "Share Alto Mayo" heading. If "list-style: none" is applied with CSS, it is recommended to add role="list" to the `<ul>` element to ensure that all screen readers report the list structure.

### Missing Programmatic Indication of Read-Only State
The "Copy Link to Share" text field is read-only, however, this is not programmatically indicated. A caret will be displayed when the field is focused, and it will be announced as a regular text field by screen readers which is confusing for all users who may attempt to edit the link text to copy the parent URL for example.

Use the "readonly" attribute to mark the `<input>` as read-only.

### Home: Incorrect List Mark-Up
In the "Landscapes" section, when the user has added several landscapes, they are encompassed in an unordered list and visually separated using a horizontal decorative line, implemented using CSS for `<li>` with class="MuiDivider-root MuiDivider-fullWidth css-1pr66se" and role="separator". This interferes with the list semantics and some screen reader users will communicate an incorrect number of items in the collection.

For example, when using NVDA with Firefox or Chrome, if three items are present in the group, the list is announced as having five items, as the two separators are included.

The same issue is observed in the "Groups" section once the user has joined several groups.

Remove the list mark-up from the decorative separators, and ensure that any element needed for their styling, such as a `<span>`, is encompassed in the previous `<li>` and is hidden from screen readers by having an aria-hidden="true" applied. The role="separator" is not needed and can be removed.

### Incorrect Heading Level in Dialog
A modal dialog is considered to be separate from the main page flow, so the modal title should be marked up as a level 1 heading. Currently, the modal title is implemented as a level 2 heading.

Decrease the heading level for the dialog title to level 1.

### Incorrect Heading Level in "Leave Landscape" Dialog
The "Leave [landscape name]" modal title is implemented as a level 2 heading instead of `<h1>` as mentioned  in a related observation, Incorrect Heading Level in Dialog.

Decrease the heading level for the dialog title to level 1.

paulschreiber:
## Description
Make share dialog and landscape markup more accessible.

### Related Issues
Fixes #495.

### DIFF
--- a/src/group/components/GroupsHomeCard.js
+++ b/src/group/components/GroupsHomeCard.js
@@ -108,7 +108,7 @@ const GroupsHomeCard = ({ groups }) => {
       </Typography>
       <List
         aria-describedby="groups-list-title"
-        sx={{ paddingLeft: theme.spacing(2), paddingRight: theme.spacing(2) }}
+        sx={{ pl: 2, pr: 2 }}
       >
         {sortedGroups.map((group, index) => (
           <React.Fragment key={group.slug}>
--- a/src/group/components/GroupsHomeCard.js
+++ b/src/group/components/GroupsHomeCard.js
@@ -31,7 +31,7 @@ const ListItem = withProps(BaseListItem, {
   component: withProps(Stack, { component: 'li' }),
 });
 
-const GroupItem = ({ group }) => {
+const GroupItem = ({ group, index }) => {
   const { t } = useTranslation();
   const navigate = useNavigate();
 
@@ -45,7 +45,15 @@ const GroupItem = ({ group }) => {
     : MEMBERSHIP_STATUS_PENDING;
 
   return (
-    <ListItem direction="column" alignItems="flex-start" spacing={2}>
+    <ListItem
+      direction="column"
+      alignItems="flex-start"
+      spacing={2}
+      sx={{
+        paddingLeft: 0,
+        borderTop: index && `1px solid ${theme.palette.gray.lite1}`, // skip first item
+      }}
+    >
       <Stack direction="row">
         <Link
           component={RouterLink}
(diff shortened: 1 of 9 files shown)


src/group/components/GroupsHomeCard.js
```diff
@@ -98,18 +106,13 @@ const GroupsHomeCard = ({ groups }) => {
       >
         {t('group.home_title')}
       </Typography>
-      <List aria-describedby="groups-list-title">
+      <List
+        aria-describedby="groups-list-title"
+        sx={{ paddingLeft: theme.spacing(2), paddingRight: theme.spacing(2) }}
```
josebui, line 111: Suggestion: You can do this inside the `sx` property and it should be the same
```suggestion
        sx={{ pl: 2, pr: 2 }}
```

josebui (approved):
Just one suggestion, that you can use in some places, it looks good

PR #504, opened 2022-08-31
fix: improve names and accessible names for group and landscape membership buttons

### LINKED ISSUE #503: Problematic names for membership leave buttons
## Non-Unique Accessible Name for Buttons
Each row in the grid has a "Connect" button. All buttons have the same accessible name of "Connect" and screen reader users that tab through the actionable elements on the page or explore the list of buttons, will not be able to differentiate between the controls. On mobile breakpoint, the table is converted into a list, and it will be harder to determine the context for each "Connect" button.

The same issue is observed for the "Member"/"Manager" buttons for landscapes that the user has joined or created. Refer as well to the related observation, Problematic Names for Buttons in Grid.

## Solution
Use the aria-label attribute to provide a descriptive name for each button e.g., "Connect: <Landscape name>". 

## Problematic Names for Buttons in Grid 
When any "Connect" button in the "Actions" column of the landscapes grid is activated, the button is replaced with a "Member" button. Similarly, landscapes that have been created by the user have a "Manager" button in the "Actions" column. When the "Member"/"Manager" buttons are activated, they open the "Leave Landscape" dialog. The purpose of these controls may not be clear for many users.

## Solution
Update the visible text of the "Member"/"Manager" buttons to "Leave" and add additional context for screen reader users as described in a previous observation. The user's role as a "Member" or "Manager" can be added in a separate column for example.

paulschreiber:
## Description
Improve names and accessible names for group and landscape membership buttons.
• Names are verbs (Leave) instead of nouns (Member/Manager)
• accessible name is unique and contains group/landscape name

### Related Issues
Fixes #503.


paulschreiber:
@josebui please review the `es` strings and translate `Leave`.

josebui (approved):

PR #513, opened 2022-09-02
fix: underline links on hover and focus

### LINKED ISSUE #512: Use of only color to convey presence of links
The presence of links (<a> with class="MuiTypography-root MuiTypography-inherit MuiLink-root MuiLink-underlineNone css-1qvyhdx") is conveyed using blue text (#307F9C), but the default underline decoration is removed via "text-decoration: none" styling. There is no alternative provided to help users identify links when they cannot perceive the color change, including on keyboard focus and mouse hover. 

Provide an additional indication of linked text that does not rely on color alone, for example by restoring the default underline decoration.

paulschreiber:
## Description
underline links on hover and focus so color isn't the only distinguishing characteristic.

### Related Issues
Fixes #512.

### DIFF
--- a/src/common/components/Table.js
+++ b/src/common/components/Table.js
@@ -80,9 +80,6 @@ const Table = props => {
         '& .MuiDataGrid-row:hover': {
           backgroundColor: 'transparent',
         },
-        '& .MuiDataGrid-row a:hover': {
-          textDecoration: 'underline',
-        },
         '.MuiDataGrid-row:nth-of-type(even)': {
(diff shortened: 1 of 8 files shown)


paulschreiber:
@josebui can we use `defaultProps` to add the underline on focus (instead of changing the .css file)?

paulschreiber:
@josebui can you push the fix using `styleOverrides` or make a GH suggestion?

josebui (approved):

PR #525, opened 2022-09-08
fix: consistent focus indicator

### LINKED ISSUE #524: Problematic focus indicator for controls
### Problem
Various focus styles are used for controls, including:
* the default outline;
* a dark teal (#307F9C) custom outline;
* a change in background color or a highlight styling that might be animated or not.

When High Contrast Mode or Contrast Themes are used in Windows environments, such focus styles that rely on a change in background color or a highlight, are not persistent or are difficult to perceive. Keyboard-only users will not be able to determine which element is currently focused. Examples include:
* The <header> links, including the highlighted options in the language selector;
* The main navigation links;
* The "Explore Landscapes"/"Connect to Another Landscape" and "Explore Groups"/"Join Another Group" links on the Home page;
* The disclosure controls for the landscape map;
* The close ("x") button for the "Share" dialog;
* The "Create Feedback" button and similar navigation controls when creating a landscape;
* The "Save Changes" and "Cancel" buttons when updating the details of a landscape;
* The "Download" button in the "Shared Files" section on the Landscape profile pages; and
* The buttons in the "Leave Landscape" modal dialog.

Additionally, in many cases the focus indicator has an animation effect of an expanding and contracting highlight with rounded sides, which lasts for as long as the element is focused. The non-essential animation can be problematic for several users, such as:
* Neurodiverse users with cognitive differences, since the movement can affect their ability to focus and process information presented on the page; and
* Users with vestibular conditions, for whom the animation may trigger symptoms such as dizziness, nausea and/or headaches.

Note as well that when the device settings are updated to indicate the user's preference for reduced motion, the focus indicator is still animated.

The change in background color or the highlighting effect is also problematic for users with low vision since it is presented with insufficient color contrast that does not meet the required 3:1 color contrast ratio against adjacent colors. Some examples are included in the following data table.

image

Additionally, when a control is focused by Pressing Tab or Shift+Tab, its text is no longer presented with sufficient color contrast, which falls below the 4.5:1 requirement for normal sized text. Some examples are included in the following data table.

image

### Solution
Implement a consistent focus style across the website that includes a persistent outline with a color contrast ratio of at least 3:1 with adjacent colors. Ensure that the text of all controls is presented with a color contrast ratio of at least 4.5:1 at all times, including the control's focused state. The animation effect of the focus indicator is not essential and should be removed.

paulschreiber:
## Description
Disable the ripple animation and use a consistent focus indicator on all controls.

### Related Issues
Fixes #524.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -30,5 +30,5 @@ code {
 
 :focus {
   outline: 2px solid #2C7690 ! important;
-  outline-offset: 1px;
+  outline-offset: 3px;
 }
--- a/src/index.css
+++ b/src/index.css
@@ -27,8 +27,3 @@ code {
 .wrap-url {
   word-break: break-word;
 }
-
-:focus {
-  outline: 2px solid #2C7690 ! important;
-  outline-offset: 3px;
-}
--- a/src/index.css
+++ b/src/index.css
@@ -27,3 +27,8 @@ code {
 .wrap-url {
   word-break: break-word;
 }
+
+:focus {
+  outline: 2px solid #2C7690 ! important;
+  outline-offset: 1px;
+}
--- a/src/theme.js
+++ b/src/theme.js
@@ -54,6 +54,14 @@ const components = {
     },
   },
   MuiButton: {
+    styleOverrides: {
+      root: {
+        '&:focus': {
+          outline: `2px solid ${colorTheme.palette.blue.dark}`,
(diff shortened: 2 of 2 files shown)


paulschreiber:
@josebui can you help translate the css changes to theme code?

josebui:
@paulschreiber I moved it, let me know how it looks

PR #530, opened 2022-09-09
fix: add aria label to map search box

### LINKED ISSUE #529: Missing Programmatic Name for Search Field
## Description
On "Create Landscape: Place your landscape on a map" page, a search field is displayed on the map with placeholder text "Enter a city, town or region". A placeholder is not sufficient as a replacement for a field label, particularly because the placeholder will no longer be available to some users of assistive technology once data is entered into the field.

### Solution
Consider adding a permanently visible label for the field using the <label> element and associate it with the <input> using the "for" attribute. The <label> can just use the same text as the existing placeholder, which should be removed. Alternatively, use the aria-label attribute to provide an accessible name for the field.

### LINKED ISSUE #531: Non-Descriptive Accessible Name for Clear Search Control
## Description
On the "Create Landscape: Place your landscape on a map" page, the search field that is provided on the map to search for a landscape location has an "x" control (<a> with class="reset") that clears the search field. This control has no accessible name and is announced as "times" by screen readers which does not describe the control purpose.

### Solution
Use the aria-label attribute to provide a descriptive accessible name for the control e.g., aria-label="Clear Search".

paulschreiber:
## Description
Add aria-label attributes to the map search box and clear search button.

### Related Issues
Fixes #529. Fixes #531.

### DIFF
--- a/src/gis/components/Map.js
+++ b/src/gis/components/Map.js
@@ -133,6 +133,9 @@ const LeafletDraw = () => {
     };
     const drawControl = new L.Control.Draw(options);
     map.addControl(drawControl);
+
+    drawControl._container.querySelector('a').setAttribute('role', 'button');
+
     return () => map.removeControl(drawControl);
   }, [editEnabled, isSmall, map, featureGroup, markerEnabled, polygonEnabled]);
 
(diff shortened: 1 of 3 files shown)


josebui (approved):

paulschreiber:
@josebui this PR is causing a couple of tests to fail. Can you help debug it?

Issue #488, opened 2022-08-19
Track Event: Visit to kobotoolbox.org from our toolbox


DerekCaelin:
We want to know when users navigate to KoBoToobox.org from the tools screen in our toolbox, either by clicking on the go-to link or the image.

<img width="588" alt="image" src="https://user-images.githubusercontent.com/10799815/185650580-86a4cbcc-78e6-4e7b-bd3d-f0bb299b58b4.png">

PR #544, opened 2022-09-28
feat: add plausible tracking to several links

david-code:
## Description

Will add plausbile tracking to a few different links. For now it is just for the Kobo Toolbox link (#488).

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


david-code:
I chose a simple solution for the first task #488 to make it so that all uses of the Tool component will be tracked as external links. I'm not sure if this is the best solution but I thought I'd start simple and see if there was any suggestions.

david-code:
Added another commit for creatings Landscapes, but it is still missing:
- country info
- if the user skips outline creation

I also think it will be triggered every time the outline is modified, I'm not sure if that is desired.

paulschreiber (commented):
This looks good. Can we also track when they click "skip"?

src/landscape/components/LandscapeForm/BoundaryStep.js
```diff
@@ -113,6 +117,9 @@ const MapDrawPolygon = props => {
       ...landscape,
       areaPolygon,
     });
+    trackEvent('Landscape created', {
```
josebui, line 120: Issue: I don't think this is necessarily creating an event after the landscape is created, it is called when the request is sent, after this still there could be backend validations, I am not sure how the function that calls the API results is propagated within that `save()` function but ideally you shoul be able to check it save was succesful using promises `save(...).then(() => track even)`

david-code: Oh yes good point.

david-code:
> 



> This looks good. Can we also track when they click "skip"?

Yep, I'll do that next.

david-code:
Added the following features in the last few commits:
- Track when a user skips adding landscape boundaries
- Track group creation
- Track user signup (this would rely on a [PR in the backend repo](https://github.com/techmatters/terraso-backend/pull/205))

david-code:
Here are the specs for map creation:

> We want to know when people create data maps. Tag event by:
>
>    Group/Landscape where created (1)
>    Dataset ID (2)
 >   Landscape of creator (if available) (3)

Here is the code that I think is already being triggered:

```javascript
// in PreviewStep.js
      const success = _.get('meta.requestStatus', data) === 'fulfilled';
      if (success) {
        trackEvent('Map: Created', {
          props: {
            owner: owner.name,
            file: visualizationConfig?.selectedFile.name,
          },
        });
        onSaved(data.payload.slug);
      }
    });
```

We have the name of the dataset and should be able to add the dataset ID, so that should be (2).

Group id is be available in the `GroupContext`, so that's 1).

I don't think I understand what (3) means. My best guess is if the map is created in a group, the landscape associated with the group should be stored as well. Any ideas?

src/landscape/components/LandscapeForm/BoundaryStep.js
```diff
@@ -285,11 +312,22 @@ const BoundaryOptions = props => {
...
   const onOptionClick = option => () => {
     option.onClick();
     scrollToNavBar();
   };
 
+  const saveWithoutBoundary = async () => {
+    await save(landscape);
+    if (props.isNew) {
+      trackEvent('Landscape created', {
+        props: { option: 'skip-boundary', country: landscape.location },
```
paulschreiber, line 326: suggestion make an `OPTION_SKIP_BOUNDARY` for consistency

paulschreiber (changes requested):

paulschreiber (approved):

PR #553, opened 2022-09-30
fix: Increased waitFor tiemout

josebui:
## Description

- Increased waitFor timeout
- Improved search field debounced value
- Fixed test scripts command


package.json
```diff
@@ -55,8 +55,8 @@
     "build": "react-scripts build",
     "build-serve": "serve -s build",
     "test": "react-scripts test --transformIgnorePatterns \"node_modules/(?!react-leaflet)/\"",
-    "test-debug": "react-scripts --inspect-brk test --runInBand --no-cache",
-    "test-coverage": "react-scripts test --coverage --changedSince=origin/main --watchAll=false",
+    "test-debug": "react-scripts --inspect-brk test --runInBand --no-cache --transformIgnorePatterns \"node_modules/(?!react-leaflet)/\"",
+    "test-coverage": "react-scripts test --coverage --changedSince=origin/main --watchAll=false --transformIgnorePatterns \"node_modules/(?!react-leaflet)/\"",
```
paulschreiber, line 59: question: should leaflet stay excluded? Or was that just for debugging?

paulschreiber (approved):

package.json
josebui: @paulschreiber yes, this is expected, eventually if we need to test some specific visual behavior that leaflet needs to do we might have to change this, but I was not able to find a better way yo do this yet

PR #555, opened 2022-10-03
fix: Update GraphQL calls for Graphene v3 compatibility

paulschreiber:
## Description
Update GraphQL calls for Graphene v3 compatibility. Use an enum type, not a string. Requires corresponding backend change.

Release notes: https://github.com/graphql-python/graphene/wiki/v3-release-notes


josebui:
Old PR

PR #590, opened 2022-10-27
feat: Create Landscape after step 1

josebui:
## Description

- Create Landscape after step 1

### DIFF
--- a/src/landscape/components/LandscapeNew.js
+++ b/src/landscape/components/LandscapeNew.js
@@ -16,6 +16,7 @@ import { useAnalytics } from 'monitoring/analytics';
 import { useFetchData } from 'state/utils';
 
 import { fetchGroupsAutocompleteList } from 'group/groupSlice';
+import { PARTNERSHIP_STATUS_NO } from 'landscape/landscapeConstants';
 import { saveLandscape, setFormNewValues } from 'landscape/landscapeSlice';
 import {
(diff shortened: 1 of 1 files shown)


src/landscape/components/LandscapeNew.js
```diff
@@ -79,21 +82,33 @@ const LandscapeNew = () => {
...
-      navigate(`/landscapes/${landscape.slug}`);
-    }
-  }, [success, landscape, navigate, dispatch]);
+  const onUpdate = async updatedLandscape => {
+    setUpdatedLandscape(updatedLandscape);
+    return dispatch(saveLandscape(updatedLandscape)).then(data => {
+      const success = _.get('meta.requestStatus', data) === 'fulfilled';
+      if (success) {
+        navigate(`/landscapes/${landscape.slug}`);
+        trackEvent('Landscape created - Optional Steps', {
```
josebui, line 91: @david-code David, what do you think about this, this feature of creating the landscape always after step 1 kinds of changes the logic for the create event in plusible tracking the map option, let me know what do you think

josebui: In this case, it is actually updating not creating the landscape, but it is still something that the user is doing through the create landsacpe steps, so maybe it is still ok

paulschreiber (approved):

PR #594, opened 2022-11-01
fix: Keep boundary current option and remove save for now

josebui:
## Description

- Keep boundary step seelcted option when going back
- Remove save for now from key info step

### DIFF
--- a/src/landscape/components/LandscapeForm/BoundaryStep.js
+++ b/src/landscape/components/LandscapeForm/BoundaryStep.js
@@ -370,11 +370,13 @@ const getOptionComponent = option => {
 };
 
 const BoundaryStep = props => {
-  const [option, setOption] = useState(OPTION_SELECT_OPTIONS);
   const [boundingBox, setBoundingBox] = useState();
   const isMounted = useIsMounted();
-  const OptionComponent = getOptionComponent(option);
   const { landscape, onSave, setUpdatedLandscape } = props;
+  const [option, setOption] = useState(
+    landscape.boundaryOption || OPTION_SELECT_OPTIONS
+  );
+  const OptionComponent = getOptionComponent(option);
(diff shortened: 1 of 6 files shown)


src/landscape/components/LandscapeForm/BoundaryStep.js
```diff
@@ -367,11 +367,13 @@ const getOptionComponent = option => {
 };
 
 const BoundaryStep = props => {
-  const [option, setOption] = useState(OPTION_SELECT_OPTIONS);
   const [boundingBox, setBoundingBox] = useState();
   const isMounted = useIsMounted();
-  const OptionComponent = getOptionComponent(option);
   const { landscape, onSave, setUpdatedLandscape } = props;
+  const [option, setOption] = useState(
+    landscape.boundaryOption || OPTION_SELECT_OPTIONS
```
paulschreiber, line 374: question:  what does OPTION_SELECT_OPTIONS mean? That name is confusing.

paulschreiber (approved):

src/landscape/components/LandscapeForm/BoundaryStep.js
josebui: @paulschreiber That is referencing the options screen where all the available options for boundary are displayed, maybe it can be `OPTION_DISPLAY_OPTIONS`? or can you think of a better name?

paulschreiber: Maybe something like`OPTION_BOUNDARY_MENU` or `OPTION_BOUNDARY_CHOICES` if that is the intent.

josebui: I like `OPTION_BOUNDARY_CHOICES`

PR #617, opened 2022-11-24
feat: Delete landscape profile photo

david-code:
Adds a new "endpoint" to the landscapeSlice that lets the user query the backend to delete the profile image and description.

Still issues with the CSS, that can come in a later commit.

Also relies on a small change to the backend.


src/localization/locales/en-US.json
```diff
@@ -245,6 +245,11 @@
         "form_add_landscape_label": "Add Landscape",
         "form_add_profile_label": "Add Profile",
         "form_save_label": "Save Changes",
+        "profile_image_delete_label": "Delete landscape photo",
+        "profile_image_delete_confirm_title": "Delete your landscape photo?",
+        "profile_image_delete_confirm_message": "[TBD] You cannot undo this. Once it's deleted a placeholder image will be displayed instead.",
```
paulschreiber, line 250:
```suggestion
        "profile_image_delete_confirm_message": "Are you sure you wish to delete the landscape photo? This cannot be undone.",
```

src/localization/locales/en-US.json
```diff
@@ -245,6 +245,11 @@
         "form_add_landscape_label": "Add Landscape",
         "form_add_profile_label": "Add Profile",
         "form_save_label": "Save Changes",
+        "profile_image_delete_label": "Delete landscape photo",
+        "profile_image_delete_confirm_title": "Delete your landscape photo?",
```
paulschreiber, line 249:
```suggestion
        "profile_image_delete_confirm_title": "Delete the landscape photo?",
```

src/landscape/landscapeSlice.js
```diff
@@ -238,6 +248,16 @@ const landscapeSlice = createSlice({
         landscape: action.payload,
         success: true,
       },
+      profile: {
+        ...state.profile,
+        landscape: {
+          ...state.profile.landscape,
+          ..._.pick(
```
josebui, line 255: Suggestion: I think here since the `saveLandscape` could potentially be used by othere parts of the profile data, I think it should update all existing fields for the profile data not just the profileImage data, so any key that exists in the `payload`

josebui: Actually I am not sure if this is being used, because I see the `deleteProfileImage` actions here, do we need this change in the saveLandscape action?

david-code: > Actually I am not sure if this is being used, because I see the deleteProfileImage actions here, do we need this change in the saveLandscape action?

Yes you are right. I originally tried just using this endpoint but in the end I think it was cleaner to great a new endpoint.  I meant to remove this change but I must have overlooked it. I'll get rid of it.

PR #619, opened 2022-11-29
fix: Add tooltip to Delete Icon for Landscape Profile Image

david-code:
## Description

Adds a tooltip to match the spec. I added a separate section to the translations JSON because it seemed like a good idea to me, but we can easily change that.


paulschreiber:
I already fixed this with https://github.com/techmatters/terraso-web-client/commit/59c6d23f6d1fdd6a720093581ec066af2a93e824.

PR #633, opened 2022-12-12
fix: Display null columns in Mapping data preview

### LINKED ISSUE #574: Data Preview in Mapping tool is misaligned
image
image

## Description
The column headers do not align with the correct data column. Eg, the "Road" header shows up over the "Who" column in the previewer. 

## Steps To Reproduce

1. Upload a data file: https://drive.proton.me/urls/3NTVJBQYTM#sxFCs6lgW9y2
2. "Make Map"
3. Select data file
4. Examine previewer

david-code:
## Description

Previously the data visualization in creating a map would not display null columns. This is due to the somewhat weird properties of Javascript arrays.

The problem: `Sheets.utils.sheet_to_json`, when called with '{header:
1}' will leave blank spots in the array. For example: `1,,3` becomes
`Array(3)[1, <1 empty slot>, 3]`. This is a Javascript array
quirk. However the previous code iterated over an array, to these
blank columns were just "skipped".

This PR changes `Sheets.utils.sheet_to_json` to use `{header:
headers}`. Because the headers are parsed from the sheet itself, this
should preserve the order of the headers as well.

### Checklist
- [x] New tests added

### Related Issues
Fixes #574

### Verification steps

In contrast to the bug report, you can see that the null rows are now displayed as blank columns.

image

### DIFF
--- a/src/sharedData/visualization/components/VisualizationConfigForm/SetDatasetStep.js
+++ b/src/sharedData/visualization/components/VisualizationConfigForm/SetDatasetStep.js
@@ -135,13 +135,14 @@ const DatasetPreview = () => {
       }),
     [colCount, rowCount]
   );
+
   const sample = useMemo(
     () =>
       SheetsJs.utils.sheet_to_json(sheet, {
         range: sampleRange,
-        header: 1,
+        header: headers,
       }),
-    [sheet, sampleRange]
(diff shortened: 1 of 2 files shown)


josebui (approved):
Looks good

PR #648, opened 2023-01-06
chore: update react dependencies

paulschreiber:
## Description
Update react dependencies.


src/gis/components/Map.js
```diff
@@ -254,11 +254,6 @@ const LeafletSearch = () => {
       inputField.setAttribute('aria-label', t('gis.map_search_placeholder'));
       inputField.setAttribute('role', 'combobox');
 
-      const clearButton =
```
josebui, line 257: Question: Why is this not needed anymore?

paulschreiber: This was an accessibility patch I added. The leaflet search people fixed the module so it now uses a button instead of a link. As a result my patch (a) is no longer needed (b) causes an error.

david-code (approved):

PR #651, opened 2023-01-09
fix: update local development host and project name

paulschreiber:
## Description
Update local development host and project name. Avoids browser opening localhost:3000, which does not work with Google SSO.


david-code (commented):
Nice, good fix.

david-code (approved):

PR #720, opened 2023-02-21
fix: set a default overflow-wrap for popup content in leaflet map widgets

### LINKED ISSUE #554: Long text exceeds dimensions of pop-up
image

## Description
I included a long URL of a photo with as a data point in a pop-up. The text exceeded the dimensions of the popup

## Steps To Reproduce

1. Join Rocky Hill
2. Look at Link: https://app.terraso.org/landscapes/rocky-hill/map/known-bike-racks
3. Click on a point with a URL attached

## Expected behavior
Extremely long text without spaces should wrap.

## Actual behavior
Extremely long text spills over the popup onto the map.

shrouxm:
## Description
sets the default text wrapping behavior for leaflet map popup widgets to "overflow-wrap: anywhere;" (see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Text/Wrapping_Text for details). this makes sure for now that text will never extend beyond the contents' box

i found this bug also existed on e.g. the landscapes view if a landscape name contained a very long word, and is fixed there as well

i'm not sure how to test for this

we would maybe like to do something more subtle eventually but i will let someone with a more designer brain put thought into that rather than overworking my store-brand brain

### Checklist
- [ ] New tests added

### Related Issues
Fixes #554 

### Verification steps
create a landscape with a name that contains a long word and select it in the map on the landscapes page, or create a data viz map that displays a string column which contains a long word

### DIFF
--- a/src/gis/components/Map.css
+++ b/src/gis/components/Map.css
@@ -64,3 +64,7 @@
 .leaflet-disabled {
   opacity: 0.6;
 }
+
+.leaflet-popup-content-wrapper {
+  overflow-wrap: anywhere;
+}


david-code (approved):
Looks like a good change to me

PR #722, opened 2023-02-22
refactor: allow Restricted component to hide children when users do not have a permission

shrouxm:
## Description
The Restricted component previously could only restrict a component to users with a permission, now it has an argument toggle whether it restricts to users with or without a permission. The new argument is used to clean up some code duplication i introduced in LandscapeView

### Verification steps
should be no visible changes, just a transparent refactor

### DIFF
--- a/src/permissions/components/Restricted.js
+++ b/src/permissions/components/Restricted.js
@@ -26,6 +26,7 @@ const Restricted = props => {
   const {
     permission,
     resource,
+    forUnallowedUsers = false,
     FallbackComponent,
     LoadingComponent,
     children,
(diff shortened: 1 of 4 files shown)


src/permissions/components/Restricted.js
```diff
@@ -26,6 +26,7 @@ const Restricted = props => {
   const {
     permission,
     resource,
+    forUnallowedUsers = false,
```
paulschreiber, line 29: suggestion: disallowed instead of unallowed.

shrouxm: i did `forUnallowedUsers -> toDisallowedUsers` since i think that's a bit more clear phrasing (restrict for maybe sounds like restricting unallowed users from seeing the thing, instead of restricting seeing the thing to them). y is phrasing so hard

paulschreiber (approved):

PR #746, opened 2023-03-06
feat: Add Microsoft as SSO provider

david-code:
## Description

Adds Microsoft SSO provider to the frontend. Not much to do here, just follows the example of the other providers.


src/account/components/AccountLogin.js
```diff
@@ -39,6 +40,11 @@ import logo from 'assets/logo.svg';
 
 import theme from 'theme';
 
+// ref: https://mui.com/material-ui/icons/#svgicon
+const MicrosoftIcon = props => {
+  return <SvgIcon {...props}><path d="M2 3h9v9H2V3m9 19H2v-9h9v9M21 3v9h-9V3h9m0 19h-9v-9h9v9Z"/></SvgIcon>;
```
david-code, line 45: Ideally the svg would be stored in a `.svg` file and somehow loaded here when it's imported, but I couldn't get this to work. Anyone have any ideas how to do this?

paulschreiber: I found a suggestion here: https://stackoverflow.com/questions/60136470/how-to-assign-a-custom-icon-to-materialui-buttons-starticon

paulschreiber (approved):

src/account/components/AccountLogin.js
paulschreiber: please rename the file to `Microsoft.svg`.

PR #782, opened 2023-03-16
fat: Added shared data gis file formats

josebui:
## Description

- Added support to upload GIS file formats

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....


paulschreiber (changes requested):
This looks good, code-wise. Please:
• run the icons through TinyPNG
• fix the failing tests
• update the [TODO] string in en and es

src/localization/locales/en-US.json
```diff
@@ -498,6 +498,7 @@
         "delete_label": "Delete “{{name}}”",
         "download_label": "Download “{{name}}”",
         "invalid_extension": "The file extension ({{file.resourceType}}) does not match the file’s contents",
+        "invalid_extension_zip": "[TODO] Invalid shapefile",
```
paulschreiber, line 501: issue: this string does not appear to be used anywhere. should it be removed?

josebui: @paulschreiber the extension comes from the context of the error message from the api, so we can customize it like that, in this case it is useful because there is a validation for zip shapefiles, if we remove this it will give the user a generic `The file extension (.zip) does not match the file’s contents` and it is not true since it can be a valid zip but not a valid zip shapefile

josebui:
@paulschreiber I already ran the icons through TinyPNG

src/localization/locales/en-US.json
paulschreiber, line 501:
```suggestion
        "invalid_extension_zip": "The file {{name}} is a valid zip, but is not a valid ESRI shapefile.",
```

paulschreiber (changes requested):

src/localization/locales/en-US.json
josebui: @paulschreiber That is not accurate, this message could also appear if it is not a valid zip in general, so what you know here is that it is not a valid zip shapefile, this means:
1. It could be a valid zip but not a valid shape zip file
2. It could be an inavlid zip file

paulschreiber: Can we separate out case 1 and 2 and give different error messages for each?

src/localization/locales/en-US.json
```diff
@@ -498,6 +498,8 @@
         "delete_label": "Delete “{{name}}”",
         "download_label": "Download “{{name}}”",
         "invalid_extension": "The file extension ({{file.resourceType}}) does not match the file’s contents",
+        "invalid_zip": "[TODO] Invalid zip",
+        "invalid_shapefile": "[TODO] Invalid shapefile",
```
paulschreiber, line 502:
```suggestion
        "invalid_zip": "The file {{name}} is a not a valid zip.",
        "invalid_shapefile": "The file {{name}} is a valid zip, but is not a valid ESRI shapefile.",
```

paulschreiber (approved):

PR #804, opened 2023-03-27
feat: Email notifications

paulschreiber:
## Description
Add a setting to enable and disable email notifications. Handle unsubscribe links.

### DIFF
--- a/src/account/components/AccountProfile.js
+++ b/src/account/components/AccountProfile.js
@@ -19,8 +19,9 @@ import React from 'react';
 import _ from 'lodash/fp';
 import { useTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
-import { useNavigate } from 'react-router-dom';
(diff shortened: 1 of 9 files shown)


src/account/components/AccountProfile.js
```diff
@@ -157,4 +174,23 @@ const AccountProfile = () => {
   );
 };
 
+const NotificationsCheckboxes = props => {
+  const { t } = useTranslation();
+  const { field } = props;
+
+  const handleChange = event => {
+    field.onChange(event.target.checked ? 'true' : 'false');
+  };
```
josebui, line 183:
```suggestion
  const handleChange = useCallback(event => {
    field.onChange(event.target.checked ? 'true' : 'false');
  }, [field.onChange]);
```

josebui (changes requested):

josebui (approved):

PR #805, opened 2023-03-27
fix: Layers not always changing on story map transition

josebui:
## Description

The story map flow provides a way to change the layers being displayed on each chapter, through the `onChapterEnter` and `onChapterExit` properties of the chapter or the title transition. Right now the `onChapterEnter` is not always executing because there is a condition for the location change on the map that will return early if there is no need for a location change, but there is a situation where there can be layer changes without a location change.

This fix is removing the early return so that the rest of the functionality can still be executed.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/src/storyMap/components/StoryMap.js
+++ b/src/storyMap/components/StoryMap.js
@@ -373,40 +373,39 @@ const StoryMap = props => {
 
         // Check if the map is already at the transition center
         const decimalPlaces = CURRENT_LOCATION_CHECK_PRESSISION;
-        if (
+        const isInLocation =
           mapCenter.lng.toFixed(decimalPlaces) ===
             transitionCenter.lng.toFixed(decimalPlaces) &&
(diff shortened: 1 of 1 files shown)


src/storyMap/components/StoryMap.js
```diff
@@ -373,40 +373,39 @@ const StoryMap = props => {
...
         // Check if the map is already at the transition center
         const decimalPlaces = CURRENT_LOCATION_CHECK_PRESSISION;
-        if (
+        const isInLocation =
           mapCenter.lng.toFixed(decimalPlaces) ===
             transitionCenter.lng.toFixed(decimalPlaces) &&
           mapCenter.lat.toFixed(decimalPlaces) ===
-            transitionCenter.lat.toFixed(decimalPlaces)
-        ) {
-          return;
```
josebui, line 382: This is the removed return

paulschreiber (approved):

PR #856, opened 2023-04-11
chore: Added anonymous access configuration doc

josebui:
## Description

- Added doc for anonymous access configuration


docs/anonymous-routes.md
```diff
@@ -0,0 +1,20 @@
...
+      topMessage: 'storyMap.optional_auth_top_message',
+      bottomMessage: 'storyMap.optional_auth_bottom_message',
+    },
+  }),
+```
+**enabled:** is a boolean that enables or disables the anonymous route. If it is set to false, the route will be accessible only to logged in users.
+
+**topMessage:** is a string that will be displayed at the top of the route content. See `OptionalAuthTopMessage` component.
+
+**bottomMessage:** is a string that will be displayed at the bottom of the route content. See `OptionalAuthBottomMessage` component.
```
paulschreiber, line 20: 1. Add `isEmbedded`, which is for content we intend for others to put in an `iframe`.

paulschreiber (changes requested):

paulschreiber (approved):

PR #860, opened 2023-04-12
build: change create-build command to build in Makefile

shrouxm:
## Description
the README says to run the command `build` to build the app for production release, when the correct command is `create-build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
 
 Terraso web client is a React application that powers the frontend of the Terraso platform.
 
-
 ## Requirements
 
 - Docker: Version 20.10.10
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ Launches the test runner in the interactive watch mode also running the [jest-ax
 
 Lunches the test runner with coverage
 
-### `make build`
+### `make create-build`
```
paulschreiber, line 41: suggestion: update the makefile instead. The backend used `make build`.

josebui: For the web client the command is `make build-docker`

shrouxm: @josebui that is the command to build the docker image, there is also a `create-build` command in the Makefile which creates a production build of the web app

paulschreiber (approved):

Issue #838, opened 2023-04-04
Extract client API code into library


shrouxm:
Since the LandPKS mobile app will be sharing a backend with the Terraso web app, and the Terraso web app is going to host some LandPKS views, there is potential/impetus for code re-use.

Some questions:
- Should we create a re-usable Terraso API client library that the web and mobile client can both depend on? (This issue assumes yes).
- Where will that library live? (My assumption is separate repo).
- How will the library be packaged/distributed? (My assumption is we will not currently bother distributing to npm, and just depend on it via git URL).
- Should we try to only include the re-used code in the library, but keep e.g. story maps API client code in the web client since we know it'll only be used there? (My assumption is it's easier to keep it all in the same place).

Update 19/04/2023:
- We had a conversation recorded here: https://docs.google.com/document/d/13ZyGeOxfJ2avDx80tLZfqVmiQOv7-UOeXbxVjqvRJ14/edit#

We agreed on the following steps forward:

```[tasklist]
### Tasks
- [ ] https://github.com/techmatters/terraso-web-client/issues/878
- [ ] https://github.com/techmatters/terraso-web-client/issues/879
- [ ] https://github.com/techmatters/terraso-web-client/issues/883
- [ ] https://github.com/techmatters/terraso-web-client/issues/880
- [ ] https://github.com/techmatters/terraso-mobile-client/issues/27
- [ ] https://github.com/techmatters/terraso-web-client/issues/881
```

josebui:
What will the `Terraso API client library` provides? Is the intent to reuse the code defined here: https://github.com/techmatters/terraso-web-client/blob/main/src/terrasoBackend/api.js

josebui:
Did we decided the framework that will be used for the mobile app? I am not sure if we can share a library like this for a mobile and web client, I am not sure if everything is compatible there, but it depends on the framework I guess.

shrouxm:
all of the frameworks under consideration are based on React, but would either use the React Native rendering library or React with MUI as we do in the web app. my thought is that the library would be concerned with retrieving data from terraso's graphql backend and then making it available through a redux store and/or hooks for consumption by a react app. the web app and mobile client would then use that library to supply data to their views

if we are able to use React with MUI for the mobile client as well we could consider having everything in one `terraso-client` repo, but i'm not sure if that's a good idea and didn't want to bother thinking about it without having confirmed it's possible to use React with MUI for the mobile client

shrouxm:
[...]

PR #897, opened 2023-04-25
refactor: migrate account logic and necessary dependencies to TypeScript

### LINKED ISSUE #878: Migrate account logic to Typescript in a separate API folder.

shrouxm:
## Description
Migrates account logic and necessary dependencies to Typescript. I cleaned up the commits such that the first 4 commits are tooling chores and moving things around. So I suggest skimming those first and then doing a focused review on the remaining commits which are the actual code migration.

### Checklist
- [x] Clean things up.
- [x] Add documentation where any weird Typescript magic is happening.

### Related Issues
Closes #878 

### Verification steps
Code should still run and build as expected.


tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Create React App",
+
+
+  "compilerOptions": {
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "module": "esnext",
+    "target": "es2015",
```
paulschreiber, line 9: question: can we use a later version?

src/state/account/auth.ts
```diff
@@ -59,4 +63,7 @@ export const refreshToken = async () => {
   Cookies.set('atoken', atoken, COOKIES_PARAMS);
 };
 
-export const getUserEmail = () => jwt(Cookies.get('atoken')).email;
+export const getUserEmail = () => {
+  const token = getToken();
+  return token === undefined ? undefined : jwt<AccessToken>(token).email;
```
paulschreiber, line 68: question: should you check if `typeof token === 'undefined'` or is that not appropriate in TypeScript?

src/state/account/accountService.ts
```diff
@@ -50,7 +56,18 @@ export const getAuthURLs = () =>
     })
   );
 
-export const fetchProfile = (params, currentUser) => {
+type UserQuery = {
+  users: {
+    edges: {
+      node: UserFragment & UserPreferencesFragment;
```
paulschreiber, line 62: question: what does the `&` do here?

src/state/terrasoBackend/api.test.ts
```diff
@@ -15,94 +15,102 @@
...
-  global.fetch.mockRejectedValue('Test Error');
-  await expect(terrasoApi.requestGraphQL()).rejects.toEqual([
+  mockFetch.mockRejectedValue('Test Error');
+  await expect(terrasoApi.requestGraphQL('', {})).rejects.toEqual([
     'terraso_api.error_request_response',
   ]);
   expect(console.error).toHaveBeenCalledTimes(1);
   expect(rollbar.error).toHaveBeenCalledTimes(1);
 });
+// TODO: is this test testing what it's supposed to test?
```
paulschreiber, line 38: issue: make sure to resolve this TODO

src/state/utils.ts
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright © 2021-2023 Technology Matters
```
paulschreiber, line 2:
```suggestion
 * Copyright © 2023 Technology Matters
```

src/state/account/auth.ts
josebui, line 68:
```suggestion
  return token ? jwt<AccessToken>(token).email : null;
```
Question: Is there a reason to use `undefined` instead of a falsy value in this case?

src/state/store.ts
```diff
@@ -39,21 +44,27 @@ const handleAbortMiddleware = store => next => action => {
...
+  group: groupReducer,
+  landscape: landscapeReducer,
+  notifications: notificationsReducer,
+  sharedData: sharedDataReducer,
+  taxonomies: taxonomiesReducer,
+  gis: gisReducer,
+  storyMap: storyMapReducer,
+});
+
+type GetReducerType<T extends Reducer> = T extends Reducer<infer S> ? S : never;
```
david-code, line 59: comment: Cool! didn't know this could be done. Is there any cases where `GetReducerType<sometype>` would evaluate to `never`?

david-code (commented):
Really cool Garo. Thanks for arraning the commits like you did, it makes reviewing it much easier.

> I will probably recreate the PR when it's ready to move out of draft to keep the commit history clean.

Why would this be necessary?

src/state/account/auth.ts
shrouxm: i think maybe that is not even necessary in javascript since es2015?

src/state/account/accountService.ts
shrouxm: i will explain more in the intro later, but it is an intersection type operator, so a type which has the fields of both its operands

src/state/utils.ts
shrouxm: question: `utils.ts` is moved from `utils.js`, same as with `store` and several other files in this PR, just git wasn't smart enough to realize that. i'm guessing i should not update this copyright since most of the code is not from this PR?

src/account/components/AccountLogin.js
```diff
@@ -33,10 +33,9 @@ import PageHeader from 'layout/PageHeader';
 import PageLoader from 'layout/PageLoader';
 import LocalePicker from 'localization/components/LocalePicker';
 import { useAnalytics } from 'monitoring/analytics';
+import { fetchAuthURLs } from 'state/account/accountSlice';
```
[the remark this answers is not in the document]
shrouxm: yes, this is all code that i will want to move (or at least most of what is in each file). i realize the naming isn't going to make the most sense right now, but didn't want to put in the effort to come up with a better folder structure that we know we're going to rearrange again in a couple weeks. i guess it is sort of a `model` folder now rather than `state`.

src/state/account/accountService.ts
```diff
@@ -16,25 +16,31 @@
  */
 import _ from 'lodash/fp';
 
-import * as terrasoApi from 'terrasoBackend/api';
+import * as terrasoApi from 'state/terrasoBackend/api';
 import {
+  UserFragment,
```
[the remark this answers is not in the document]
shrouxm: i defined a `User` type in `accountSlice.ts`, the `UserFragment` corresponds to what is in our GraphQL API, and the `User` type is the representation in the redux store. In general I think it's good to decouple these, and right now they are in fact different types (`UserFragment` has edges and nodes, `User` does not since they are flattened). `UserFragment` is a type that could be auto-generated if we decide to set up code generation on our GraphQL schema and queries

not sure if it will help with the confusion, but the standard naming convention in Typescript is that types are capitalized and values are lower-cased. So `xFragment` is a graphql fragment, and `XFragment` is the Typescript type of the corresponding object that would be created by including that fragment in a query. is there a different naming convention that would make more sense?

src/state/utils.ts
paulschreiber: Use 2023 for newly created-files.

src/state/account/accountService.ts
```diff
@@ -65,44 +82,30 @@ export const fetchProfile = (params, currentUser) => {
...
-    .then(user => ({
-      ..._.omit('preferences', user),
-      preferences: parsePreferences(user),
-    }));
+  const result = await terrasoApi.requestGraphQL<UserQuery>(query, {
+    email: currentUser?.email,
+  });
+
+  const user = result.users.edges.at(0);
+  if (user === undefined) {
```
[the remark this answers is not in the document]
shrouxm: they are falsy, i just have a style preference for explicitly checking for equality to `undefined` so it's clear what is actually being checked for. Typescript will raise an IDE/compile error if you try to do anything with a `null` or `undefined` type, and will also raise a warning if you do an unnecessary check. so in this case it can actually be concluded from the fact that this all type checks that the `at` function can return undefined values but not null values (`at` doesn't raise an exception if you access something that doesn't exist, just returns undefined).

the logic is changing because Typescript is not very good at typing big lodash expressions. it can often do it, and i tried to keep with lodash where possible to keep the PR smaller, but in this case it was cleaner to just break it out. i do think this code is easier to understand for someone new to the codebase, but i know there's always a tradeoff with lodash's succinctness/robustness. i do think that with Typescript there's less of a tradeoff since the types can provide the robustness, i.e. in a sense both Typescript and lodash are smart enough to handle the null and undefined values, just one does it implicitly and one explicitly

src/state/account/accountService.ts
```diff
@@ -65,44 +82,30 @@ export const fetchProfile = (params, currentUser) => {
...
-    .requestGraphQL(query, { email: getUserEmail() })
-    .then(_.get('users.edges[0].node'))
-    .then(user => user || Promise.reject('not_found'))
-    .then(user => ({
-      ..._.omit('preferences', user),
-      preferences: parsePreferences(user),
-    }));
+export const fetchUser = async () => {
+  const email = getUserEmail();
+  return fetchProfile(null, email === undefined ? null : { email });
```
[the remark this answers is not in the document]
shrouxm: it is sort of a Typescript thing, since in Typescript you don't need to worry about forgetting to check for null (the compiler will tell you if you do). but it's also my style preference which i don't feel strongly about so i can make these checks shorter since it seems they also looked strange to paul

src/state/account/auth.ts
shrouxm: answered above, just style preference

src/account/components/Unsubscribe.js
```diff
@@ -29,10 +29,9 @@ import PageContainer from 'layout/PageContainer';
 import PageLoader from 'layout/PageLoader';
 import { useAnalytics } from 'monitoring/analytics';
 import { addMessage } from 'notifications/notificationsSlice';
+import { unsubscribeFromNotifications } from 'state/account/accountSlice';
```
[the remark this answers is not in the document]
shrouxm: i think so yeah. i don't question the formatter i just exist to comply to its demands.

src/state/store.ts
shrouxm: no, since you have a constraint that T must extend reducer on the lefthand side. IMO the type-level language is weird and gross and should be avoided at all costs, but i don't think there's any choice here.

shrouxm:
@david-code re re-creating the PR: i didn't think anyone was going to review yet, so i wanted to keep the nice commit order to make it easier to review even as i was tweaking. but i think it doesn't matter now

src/state/account/accountService.ts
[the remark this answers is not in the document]
josebui: That makes sense if potentially, we can even create the GraphQL fragments from those types then this makes sense

src/account/components/AccountLogin.js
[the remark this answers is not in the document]
josebui: `model` is still confusing, I think this is related to the api interaction in general right? I think `terrasoApi` or something related to the api might make sense for now

src/state/utils.ts
shrouxm: i am still confused, should i also then use 2023 for `store.ts`, `accountSlice.ts`, etc? i think some of those files haven't change contents at all other than the extension

src/state/account/accountService.ts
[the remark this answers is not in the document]
shrouxm: it would go the other way (creating types from GraphQL fragments) but yeah we could avoid the code dup

shrouxm:
@paulschreiber @josebui I moved this out of draft state. Not everything is perfectly well typed (especially `api.ts`), but I think that's OK for now since chasing the long tail would probably end up requiring migrating a bunch more code & keep making this initial PR bigger. I can continue refining the types as I get to know the codebase a bit better while working through the remaining migration steps outlined here: https://github.com/techmatters/terraso-web-client/issues/838

Is there anything else you'd like to see from this PR before merging?

josebui:
@shrouxm I agree, I think this si functionally enough, the only thing I would like us to improve before merging is the `state` folder name `state` inside a repo that uses redux and react is really confusing even for a temporary solution

PR #928, opened 2023-05-10
fix: inconsistent text validating account profile changes

jodiezhuu:
## Description
Inconsistent text when validating account profile changes

## Expected behaviour
- [X] The text should say "Enter a given name."
- [X] Spanish text changed to corresponding text ("given name" instead of "firstname")

## Verification steps
1. Go to URL: https://app.terraso.org/account/profile
2. Clear the given name field
3. Press save profile
4. Observe text
Screen Shot 2023-05-10 at 12 20 56 PM
Screen Shot 2023-05-10 at 1 40 21 PM

### DIFF
--- a/src/account/components/AccountProfile.js
+++ b/src/account/components/AccountProfile.js
@@ -39,7 +39,11 @@ import AccountAvatar from './AccountAvatar';
 
 const VALIDATION_SCHEMA = yup
   .object({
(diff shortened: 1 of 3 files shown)


david-code (approved):
Nice Jodie, this looks exactly right. I didn't about the localization in the validation schema myself, so I learned something new too!

PR #934, opened 2023-05-11
refactor: migrate group logic to typescript

### LINKED ISSUE #879: Migrate group membership logic to the API folder.

shrouxm:
## Description
I suggest reviewing like so:
- adding codegen: https://github.com/techmatters/terraso-web-client/pull/934/commits/88fc792814baee07ac8d2c1ef8124ae829a63f17
- changing the existing TS code to use the codegen: https://github.com/techmatters/terraso-web-client/pull/934/files/88fc792814baee07ac8d2c1ef8124ae829a63f17..ced7a83fd5ae10c2ae234fd692713c16c27c4ce1
- moving lots of files around: https://github.com/techmatters/terraso-web-client/pull/934/commits/3d1a79e1e932c91ae1aead358e04063e17894655
- actual migration of group logic to typescript: https://github.com/techmatters/terraso-web-client/pull/934/commits/e8e032e9e6eacb834bfe7d18fd37af3b45ebf11b

Note that I'm using the graphql schema from here: https://github.com/techmatters/terraso-backend/pull/481. The types would be a bit less complicated if I kept pushing that PR farther but I don't want to do too much at once. An implication of this is that some of the service code is no longer handling null/missing values that the server itself never returned, but which our own client code (especially tests) did return, so I had to make changes in a couple places.

### Checklist
- [x] relevant backend code has been merged
  - [x] https://github.com/techmatters/terraso-backend/pull/479
  - [x] https://github.com/techmatters/terraso-backend/pull/481
- [x] update README for new dev workflow

Follow-up issues to file:
- [x] figure out how to use generated types to help with mocking requests (i uncovered one test that was never doing anything, there could be more)
- [x] further simplify code by identifying more areas where API surfaces are too broad
- [x] figure out a better way of handling our server's response error format. (queries can either have an HTTP failure, or the body will have an `errors` field instead of the expected body, and mutations are the same but can additionally have `errors` fields embedded further into the request). i would either like to have a cleaner way of modeling that on the client side (maybe with separate functions for queries/mutations), or figure out how big a lift it would be to simplify the error format the server returns.

### Related Issues
Closes #879 

### Verification steps
Everything should still work exactly the same.


src/terrasoApi/group/groupUtils.ts
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright © 2021-2023 Technology Matters
```
paulschreiber, line 2:
```suggestion
 * Copyright © 2023 Technology Matters
```

package.json
```diff
@@ -57,7 +57,8 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "lint-js": "eslint ./ --fix --ext .js --ext .ts --ext .tsx --max-warnings 0 --ignore-pattern node_modules/ --ignore-pattern build/",
+    "generate": "npx graphql-code-generator",
```
josebui, line 60: Question: Will this affect the build or start command?

src/terrasoApi/schema.graphql
```diff
@@ -0,0 +1,939 @@
+schema {
```
josebui, line 1: Question: Was this file generated automatically? If it wasn't is there a way to split it so that we can organize better, maybe even using the apps categorization that we have in the backend

src/terrasoApi/user/userFragments.ts
```diff
@@ -15,7 +15,7 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-export const userFields = `
+export const userFields = /* GraphQL */ `
```
josebui, line 18: Question: What is the `/* GraphQL */` doing?

src/terrasoApi/account/accountService.ts
```diff
@@ -15,27 +15,24 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 import _ from 'lodash/fp';
-import * as terrasoApi from 'terrasoApi/terrasoBackend/api';
-import {
-  UserFragment,
-  UserPreferenceNode,
+import { graphql } from 'terrasoApi/gql';
```
josebui, line 18: Question: I don't see this file in this code, is this generated by the `generate` command? If that is it, should we hook the generate command to other commands that need this file?

src/terrasoApi/user/userFragments.ts
```diff
@@ -60,9 +44,3 @@ export const userPreferences = `
   }
   ${userPreferencesFields}
 `;
-
-export type UserPreferencesFragment = {
```
josebui, line 64: Question: I see that this type is deleted, but it is still imported in accountService.js, how is that working?

src/terrasoApi/account/accountService.ts
```diff
@@ -55,20 +52,12 @@ export const getAuthURLs = () =>
...
-};
-
 export const fetchProfile = async (
-  params: any,
+  params: null,
   currentUser: { email: string } | null
 ) => {
-  const query = `
-    query user($email: String!){
+  const query = graphql(`
```
josebui, line 59: Question: how does the `graphql` function knows what fragments to add?

josebui, line 59: Question: What is the `graphql` function doing? is doing a validation? more than that? or something different?

src/terrasoApi/schema.graphql
shrouxm: yes, the code which generates it is in the backend and was merged here: https://github.com/techmatters/terraso-backend/pull/479, although this particular schema is generated by the backend code in https://github.com/techmatters/terraso-backend/pull/481

i'm not sure the best way to handle this, if we should have the file checked into both repos, or include the one from the backend repo with a git reference in package.json or a git submodule or some such. i defaulted to checking it in because it was the most straightforward. i also don't know if it could be split up, but the code to do so would be in the backend and i think should be a future PR, but i can file an issue in the backend if that seems worth putting the effort in

package.json
shrouxm: it needs to be run any time a tagged fragment or the big schema file changes for local tooling to work, although all the code that gets generated is .gitignore'd so it doesn't matter for committing code. you can leave `npm run generate -- --watch` running in the background and it will work automagically. 

i wasn't sure what people's dev workflows look like so i didn't want to make opinionated changes to the commands, but we could consider including it as a pre-step in other commands

src/terrasoApi/user/userFragments.ts
shrouxm: it is being generated automatically because of the `/* GraphQL */` comment

src/terrasoApi/user/userFragments.ts
shrouxm: it marks the code for inclusion in the graphql validation & type generation, i will explain more in another comment

src/terrasoApi/account/accountService.ts
shrouxm: ok here is a full explanation of how the code generation works:

- by parsing `src/terrasoApi/schema.graphql`, the code gen understand which types exist in our schema, which queries & mutations are allowed, and what the types of their results are
- each _fragment_ tagged with a `/* GraphQL */` comment is validated to ensure all the fields it references exist, and a corresponding fragment type is created
- each query or mutation tagged with a ``graphql(` ... `)`` call is validated to ensure all fields and inputs exist, and an entry is created in a generated dictionary where the keys are the literal strings in the `graphql` call, and the values are the query strings with fragment spreads expanded (at code generation time) from fragments ingested via `/* GraphQL */` tags. the expanded query strings are also wrapped with the type information of the inputs and returned fields (which our `requestGraphQL` function uses to provide static typing for input and result JSON objects).
- a function `graphql` is generated which:
  - at compile time, knows what the corresponding input and result types are based on the literal string argument
  - at runtime, simply looks up its argument in the generated dictionary

if you run `npm run generate` in a local project, you can see the code which gets generated under `src/terrasoApi/gql/...`

some implications:
- you no longer need to write the fragment spread name twice. i think there is a way to set up your local tooling so that jump to definition works on the graphql string itself
- the graphql name of fragments and queries (the name you give them within the graphql string) must now be globally unique. (this is a bit annoying but i think overall worth it).
- this doesn't affect existing code, which can still operate as normal.
- you can't build queries programmatically via string interpolation using this method, the only way to re-use query code is via fragments and inputs. there are features available for type-safe programmatic manipulation of queries if we want to do that, but it's more complicated and wasn't needed for this PR so i left it out
- performance: i don't think significant difference, we aren't doing runtime string interpolation but we are doing a runtime dictionary lookup. i guess it's a bit faster, but i doubt this is a bottleneck either way

src/terrasoApi/account/accountService.ts
shrouxm: yes, & the `generate` command needs to run for the client to build at all because of type checking. see https://github.com/techmatters/terraso-web-client/pull/934#discussion_r1195464222 for codegen discussion, https://github.com/techmatters/terraso-web-client/pull/934#discussion_r1195425986 for discussion of how to set up tooling

src/terrasoApi/account/accountService.ts
shrouxm: at code generation time, the argument strings to the `graphql` function become keys in a dictionary where the values are the query with the fragments included (the fragments are also parsed at code generation time, must be tagged with `/* GraphQL */`. see https://github.com/techmatters/terraso-web-client/pull/934#discussion_r1195464222

package.json
josebui: I think if this is needed to run start the application it should be part of the `start` script or maybe if there is a better way to do handle this maybe defined in the README file, but I think ideally we want the same way the application automatically changes with code changes this files should be updated automatically without executing a separate script

src/terrasoApi/schema.graphql
josebui: So, if there is another client that needs this file, what is the process to get it and maintain it? I don't see a version referrence in this file, how can you identify if it has changed? or if the changes here are up to date?

src/terrasoApi/account/accountService.ts
josebui: The implications look managable for me, the big one related to the development environment and being able to just add changes without running a separate script everytime I think is important IMO. The one related to string interpolation I think is fine, ideally we should pass values through variables that as far as I see this is still supported.

package.json
shrouxm: yeah i agree, i'm just not sure how to do it but i'll fiddle for a bit. the problem is needing to start two background processes, & then making sure that if you stop one you stop the other & that their outputs don't clash. and the most straightforward way is to just start them both in separate terminals manually, even though it's annoying

src/terrasoApi/schema.graphql
shrouxm: re: if there's another client that needs the file, the eventual plan is for the file to live in the shared client repo

re: version reference, this is a problem but i think it was already a problem, if you make a breaking backend change there was no way before to automatically know that you need to update the frontend. and if it's a non-breaking backend change then the schemas should be compatible and so it's ok if the schema falls out of date. if you try to query a new field while the schema is out of date the types will complain and you can update the schema

it might be easier to figure out the best way to handle this by doing it for a bit. manually updating it seems annoying but at the same time the only reason you would need to manually update it is because you need to make changes to the client anyway so maybe it's not that big a deal?

shrouxm:
@josebui from my understanding, the outstanding conversations are:

- tooling: i think my latest change is a better compromise here. i added the generation command before all the relevant package scripts (start, testing, build), and it has a silent watch mode for start and test so it will keep running in the background. if someone develops without start or test running in the background but still want the graphql types to stay up to date they will need to manually call generate, i will add a line about this in the README but it seems more of an edge case. you can't see the output of the code gen command any more, but it only tells you info if there's an error in your graphql code and if you set up the graphql language server then you will see that inline anyway, which is why i think it's the best compromise
- versioning/including the schema file: i think it is straightforward to include a direct reference to the schema via a devDependency git commit reference to the backend repo (once this is merged: https://github.com/techmatters/terraso-backend/pull/507), so i think that is the cleanest way to do this. we could use tags on the backend to try to make it clearer what the commit hashes mean as a future improvement

let me know if those seem like ok solutions for the issues you raised!

src/terrasoApi/account/accountService.ts
shrouxm: see https://github.com/techmatters/terraso-web-client/pull/934#issuecomment-1552260099 re dev environment

src/terrasoApi/schema.graphql
shrouxm: ok i think i have a better solve, see https://github.com/techmatters/terraso-web-client/pull/934#issuecomment-1552260099 for details

package.json
shrouxm: i think i have a better compromise, see https://github.com/techmatters/terraso-web-client/pull/934#issuecomment-1552260099 for details

josebui (approved):
@shrouxm Thanks Garo, those 2 improvements are good solutions, eventually for versioning with tags it will be better, but for now that is giving us a clear indication if we are up to date

PR #936, opened 2023-05-12
fix: API, web client URL validation inconsistent

jodiezhuu:
## Description
API and web client URL validation are inconsistent

### Checklist
- [X] Web client identifies this URL as invalid

### Steps to Reproduce
1. Go to URL: [http://127.0.0.1:3000/landscapes/[landscape](http://127.0.0.1:3000/landscapes/%5Blandscape) slug]/upload
2. Go to Share Links tab
3. Add a link with test.c as the URL
4. Click Share Files and Links
5. See API error

Screen Shot 2023-05-12 at 3 48 59 PM


paulschreiber (changes requested):
@jodiezhuu Let's avoid rewriting the whole URL validation function — it's complicated and we don't want to maintain it.

How about this
- file a bug with yup (and submit a PR if we can figure out their code)
- add a custom validation that checks for single-letter TLDs after we run it through yup's existing validator

jodiezhuu:
> @jodiezhuu Let's avoid rewriting the whole URL validation function — it's complicated and we don't want to maintain it.
> 
> How about this
> 
> * file a bug with yup (and submit a PR if we can figure out their code)
> * add a custom validation that checks for single-letter TLDs after we run it through yup's existing validator

noted and updated

src/forms/yup.js
```diff
@@ -19,6 +19,28 @@ import * as yup from 'yup';
 
 const ARRAY_INDEX_REGEX = /\[([^)]+)\]/;
 
+const hasValidTLD = urlString => {
+  try {
+    const TLD_REGEX = /\.([^.]+)$/;
```
paulschreiber, line 24: issue: this allows any character. TLDs need to end in two or more letters. Perhaps something like `\.[a-z]{2,}$`?

src/sharedData/components/SharedDataUpload/ShareDataLinks.js
```diff
@@ -54,7 +54,14 @@ import { groupDataEntryUploadsByStatus } from './utils';
 
 export const VALIDATION_SCHEMA = yup
   .object({
-    url: yup.string().trim().ensure().transform(transformURL).url().required(),
+    url: yup
+      .string()
+      .trim()
+      .ensure()
+      .transform(transformURL)
+      .validTld()
```
paulschreiber, line 62: issue: please add `validTld()` to the other places we call transformURL (GroupForm, KeyInfoStep, RichTextEditor/index.js)

paulschreiber (changes requested):

src/forms/yup.js
```diff
@@ -19,6 +19,28 @@ import * as yup from 'yup';
...
 
+const hasValidTLD = urlString => {
+  try {
+    const TLD_REGEX = /\.[a-z]{2,}$/i;
+    const url = new URL(urlString);
+    const match = url.hostname.match(TLD_REGEX);
+    if (match) {
+      return true;
+    }
+    return false;
```
josebui, line 30: Suggestion: I think this should do the same
```suggestion
    return !!match;
```

src/forms/yup.js
```diff
@@ -19,6 +19,28 @@ import * as yup from 'yup';
...
+  } catch (e) {
+    return false;
+  }
+};
+
+yup.addMethod(yup.string, 'validTld', function () {
+  return this.test(
+    'validTld',
+    params => ({ key: 'form.validation_url_invalid', params }),
+    value => (!value ? true : hasValidTLD(value))
```
josebui, line 40: Question: Why is empty valid?

src/forms/yup.js
paulschreiber: I recommend `Boolean(match)` instead of `!!match`.

src/common/components/RichTextEditor/index.js
```diff
@@ -104,7 +104,7 @@ const withInlines = editor => {
 
   editor.insertText = text => {
     if (text && isUrl(text)) {
-      wrapLinkPartial(editor)(transformURL(text), text);
+      wrapLinkPartial(editor)(transformURL(text).validTld(), text);
```
josebui, line 107: Question: how is that validTld() function executing?, as far as I can see transformURL() returns a string

jodiezhuu: I believe you're right. I thought it was validating the input, but looking at it now, I don't think the validTld() is necessary.

paulschreiber (approved):

PR #943, opened 2023-05-15
fis: Plausible story map events update

josebui:
## Description

Story map events added and updated:
- `storymap.start`: Added when a user enters `/tools/story-maps/new`
- `storymap.publish`: When a story map changes from draft to published
- `storymap.save`: When a story map is updated after it gets published
- `storymap.saveDraft`: When a story map gets saved and it is not published

### Checklist
- [ ] New tests added

### Related Issues
Fixes https://github.com/techmatters/terraso-product/issues/94

### DIFF
--- a/src/storyMap/components/StoryMapUpdate.js
+++ b/src/storyMap/components/StoryMapUpdate.js
@@ -64,14 +64,22 @@ const StoryMapUpdate = props => {
 
     const { title, slug, storyMapId, published } = saved;
     setSaved(null);
+    const url = generateStoryMapUrl({ slug, storyMapId });
+
+    console.log({ published, storyMap });
+    const event = published
+      ? storyMap.isPublished
+        ? 'storymap.save'
+        : 'storymap.publish'
+      : 'storymap.saveDraft';
+
+    trackEvent(event, {
+      props: {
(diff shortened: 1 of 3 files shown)


src/storyMap/components/StoryMapUpdate.js
```diff
@@ -64,14 +64,22 @@ const StoryMapUpdate = props => {
 
     const { title, slug, storyMapId, published } = saved;
     setSaved(null);
+    const url = generateStoryMapUrl({ slug, storyMapId });
+
+    console.log({ published, storyMap });
+    const event = published
+      ? storyMap.isPublished
+        ? 'storymap.save'
```
paulschreiber, line 72: question: would update or revise make more sense?
```suggestion
        ? 'storymap.update'
```

paulschreiber (commented):
Don't forget to update the plausible settings page with these new goals. Will send you the script.

src/storyMap/components/StoryMapUpdate.js
josebui: @paulschreiber I will tag you in the slack thread so that you can coordiante this with Derek

paulschreiber (approved):

PR #956, opened 2023-05-20
Fix: customize message for blank select values

### LINKED ISSUE #638: Customize message for blank select values in Forms
## Description
- Customize message for blank select values in Forms
- Right now all fields in forms use the same error message if blank

## Steps To Reproduce

1. Go to URL: https://app.staging.terraso.net/landscapes/[landscape slug]/edit
2. Clear value in country select
3.  Watch error message

## Expected behavior
The message for empty Select should be different than other fields

## Actual behavior
The same message appear for all fields when the value is required

## Additional context

image

jodiezhuu:
## Description
yup.setLocale section is modified to include a condition based on the path parameter in the required method

### Checklist
- [X] Customize message for blank select values in Forms

### Related Issues
Fixes #638 

### Verification steps
1. Go to URL: https://app.staging.terraso.net/landscapes/[landscape slug]/edit
2. Clear value in country select
3. Watch error message

### DIFF
--- a/src/forms/yup.js
+++ b/src/forms/yup.js
@@ -19,16 +19,19 @@ import * as yup from 'yup';
 
 const ARRAY_INDEX_REGEX = /\[([^)]+)\]/;
 
+yup.addMethod(yup.string, 'selected', function () {
+  return this.required(params => ({
+    key: 'form.validation_select_field_required',
+    params,
+  }));
+});
+
 // Localization codes form Yup schema validation
 // Check: https://github.com/jquense/yup#api to know the format to add more codes here
 yup.setLocale({
   mixed: {
     default: params => ({ key: 'form.validation_field_invalid', params }),
-    required: params => {
-      return params.path === 'location'
-        ? { key: 'form.validation_select_field_required', params }
-        : { key: 'form.validation_field_required', params };
-    },
+    required: params => ({ key: 'form.validation_field_required', params }),
     notType: params => ({ key: 'form.validation_field_noType', params }),
   },
   string: {
--- a/src/forms/yup.js
+++ b/src/forms/yup.js
@@ -24,7 +24,11 @@ const ARRAY_INDEX_REGEX = /\[([^)]+)\]/;
 yup.setLocale({
   mixed: {
     default: params => ({ key: 'form.validation_field_invalid', params }),
-    required: params => ({ key: 'form.validation_field_required', params }),
+    required: params => {
+      return params.path === 'location'
+        ? { key: 'form.validation_select_field_required', params }
+        : { key: 'form.validation_field_required', params };
+    },
     notType: params => ({ key: 'form.validation_field_noType', params }),
   },
   string: {
--- a/src/landscape/components/LandscapeForm/KeyInfoStep.js
+++ b/src/landscape/components/LandscapeForm/KeyInfoStep.js
@@ -38,7 +38,7 @@ const VALIDATION_SCHEMA = yup
   .object({
     name: yup.string().trim().required(),
     description: yup.string().max(MAX_DESCRIPTION_LENGTH).trim().required(),
-    location: yup.string().trim().required(),
+    location: yup.string().trim().selected(),
     email: yup.string().trim().email(),
     website: yup.string().trim().ensure().transform(transformURL).url(),
   })
--- a/src/localization/locales/en-US.json
+++ b/src/localization/locales/en-US.json
@@ -702,6 +702,7 @@
         "validation_field_invalid_unexpected": "Field must be valid",
         "validation_email_invalid": "Enter a valid email address.",
         "validation_field_required": "Enter a {{path, lowercase}}.",
+        "validation_select_field_required": "Select a {{path, lowercase}}.",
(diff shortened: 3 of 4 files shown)


jodiezhuu:
Not sure if this is the best solution for #638;

Another possible solution: Introduce a `selected` localization method, which is similar to `required` but specifically for selection/dropdown components; an advantage of this could be that the validation schema is more organized and we could avoid adding more conditions to `required` in the future (if we wish to customize an error message for another component). 

Would love to hear any other ideas!

src/forms/yup.js
```diff
@@ -24,7 +24,11 @@ const ARRAY_INDEX_REGEX = /\[([^)]+)\]/;
 yup.setLocale({
   mixed: {
     default: params => ({ key: 'form.validation_field_invalid', params }),
-    required: params => ({ key: 'form.validation_field_required', params }),
+    required: params => {
+      return params.path === 'location'
```
paulschreiber, line 28: question: does this only work for the field named `location`? We want it to work for any <select>.

josebui, line 28: Issue: `yup.js` ideally should not be aware of what `location` is. I think for this you should be able to define a custom required message for the specific validation:
https://github.com/techmatters/terraso-web-client/blob/main/src/landscape/components/LandscapeForm/KeyInfoStep.js#LL41C23-L41C46
```
location: yup.string().trim().required(),
```
Maybe something like this:
```
location: yup.string().trim().required(params => ({ key: 'form.validation_select_field_required', params })),
```

josebui (changes requested):

src/forms/yup.js
jodiezhuu: Just added a custom validation rule just for selected that can be added to any validation schema

src/forms/yup.js
```diff
@@ -19,16 +19,27 @@ import * as yup from 'yup';
...
+  return true;
+};
+
+yup.addMethod(yup.string, 'selected', function () {
+  return this.test(
+    'selected',
+    params => ({ key: 'form.validation_select_field_required', params }),
+    value => isCountrySelected(value)
+  );
+});
```
paulschreiber, line 35: @josebui would it make sense to use an arrow function here?

src/forms/yup.js
```diff
@@ -19,16 +19,27 @@ import * as yup from 'yup';
 
 const ARRAY_INDEX_REGEX = /\[([^)]+)\]/;
 
+const isCountrySelected = value => {
+  if (!value || value === '') {
+    return false;
+  }
+  return true;
+};
```
paulschreiber, line 27: suggestion: remove this function

src/forms/yup.js
```diff
@@ -19,16 +19,27 @@ import * as yup from 'yup';
...
+
+yup.addMethod(yup.string, 'selected', function () {
+  return this.test(
+    'selected',
+    params => ({ key: 'form.validation_select_field_required', params }),
+    value => isCountrySelected(value)
```
paulschreiber, line 33: suggestion: Use something like `value => Boolean(value)`

paulschreiber (changes requested):

src/forms/yup.js
jodiezhuu: Thanks! Didn't know that was a built in function

src/landscape/components/LandscapeForm/KeyInfoStep.js
```diff
@@ -38,7 +38,7 @@ const VALIDATION_SCHEMA = yup
   .object({
     name: yup.string().trim().required(),
     description: yup.string().max(MAX_DESCRIPTION_LENGTH).trim().required(),
-    location: yup.string().trim().required(),
+    location: yup.string().trim().selected().required(),
```
josebui, line 41: Question: Do we need selected() and required() too? it seems that selected is a customized required check

src/forms/yup.js
```diff
@@ -19,6 +19,14 @@ import * as yup from 'yup';
 
 const ARRAY_INDEX_REGEX = /\[([^)]+)\]/;
 
+yup.addMethod(yup.string, 'selected', function () {
+  return this.test(
+    'selected',
+    params => ({ key: 'form.validation_select_field_required', params }),
+    value => Boolean(value)
```
josebui, line 26: Suggestion: I don't think we need to define a new way to check if something is required, I think you should be able to use required() inside this and define the custom message, something like: 
```javascript
yup.addMethod(yup.string, 'selected', function () {
  return this.required(params => ({
    key: 'form.validation_select_field_required',
    params,
  }));
});
```

src/forms/yup.js
josebui: @paulschreiber you mean instead of `'selected', function () {`? If you see there yup expects the use of `this` inside it, so we have to use `function() {}` to be able to use that binded this

josebui (approved):

paulschreiber (approved):

paulschreiber:
@jodiezhuu Please look over the forms and see if other pages also need to be updated to use the new `selected()` method.

PR #964, opened 2023-05-24
fix: use RouterButton instead of RouterLink for Update Maps button

paulschreiber:
## Description
The Update Map button had underline on focus because it was also a RouterLink. I removed the RouterLink component and added an onClick handler (which matches other buttons in the app).

### DIFF
--- a/src/landscape/components/LandscapeView.js
+++ b/src/landscape/components/LandscapeView.js
@@ -360,8 +360,9 @@ const LandscapeView = () => {
                 <CardActions sx={{ paddingTop: 0 }}>
                   <Button
                     variant="outlined"
-                    component={RouterLink}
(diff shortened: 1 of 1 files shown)


src/landscape/components/LandscapeView.js
```diff
@@ -360,8 +360,9 @@ const LandscapeView = () => {
                 <CardActions sx={{ paddingTop: 0 }}>
                   <Button
                     variant="outlined"
-                    component={RouterLink}
-                    to={`/landscapes/${landscape.slug}/boundaries`}
+                    onClick={() =>
```
josebui, line 363: Issue: There is a component called `RouterButton` ideally that should be used here

josebui (changes requested):

PR #966, opened 2023-05-26
feat: isolate shared API code from non-shared code

### LINKED ISSUE #883: Abstract web client specific dependencies out of API folder.

shrouxm:
## Description
This PR does a final pass to separate out code that will be migrated to the shared client codebase. All code to be migrated lives under `src/terrasoApi/shared`, and should only import code that also lives under `src/terrasoApi/shared`, with the hope that when it gets moved to another repo this repo will just need to do a text replace of imports from `'terrasoApi/shared'` to whatever the package gets named.

Some parts I am looking for particular feedback on:
- The way I have set up config settings, since shared code cannot rely on a universal way to get config settings from the environment. See the changes in the two `config.ts` files (one is new, one renamed from `config.js`).
- The way I have split out the group code into a non-shared `groups{Service,Slice,etc}` and a shared `memberships{Service,Slice,etc}`.

All but the last two commits are moving things around, and then the last 2 (and especially the last commit) have functional changes.

(@jodiezhuu this is a pretty big PR, I requested review in case you wanted to take a look & had any comments but don't feel pressure to read all of it or read it carefully if you have better things to do with your time!)

### Checklist
- [ ] manual QA reveals no errors

### Related Issues
Closes #883

### Verification steps
Everything should work exactly the same.


src/config.ts
```diff
@@ -28,8 +32,21 @@ export const GRAPH_QL_ENDPOINT = 'graphql/';
 export const COOKIES_DOMAIN =
   process.env.REACT_APP_COOKIES_DOMAIN || '127.0.0.1';
 
+const COOKIES_PARAMS = { path: '/', domain: COOKIES_DOMAIN };
+
 export const ROLLBAR_TOKEN = process.env.REACT_APP_ROLLBAR_TOKEN;
 
+setAPIConfig({
+  terrasoAPIURL: TERRASO_API_URL,
+  graphQLEndpoint: GRAPH_QL_ENDPOINT,
```
paulschreiber, line 41:
```suggestion
  graphQLEndpoint: GRAPHQL_ENDPOINT,
```

paulschreiber: Update definition of `GRAPHQL_ENDPOINT` above as well.

paulschreiber (changes requested):

src/terrasoApi/shared/memberships/membershipsFragments.ts
```diff
@@ -83,6 +25,10 @@ export const groupMembersPending = /* GraphQL */ `
 
 export const groupMembersInfo = /* GraphQL */ `
   fragment groupMembersInfo on GroupNode {
+    id
```
josebui, line 28: Question: Do we need to add these group fields here? I always see this fragment used next to the groupFields fragment, or where is this needed?

src/terrasoApi/store.ts
```diff
@@ -14,59 +14,38 @@
...
+export type AppDispatch = DispatchFromStoreFactory<typeof createStore>;
 
-const reducer = combineReducers({
-  account: accountReducer,
-  userHome: userHomeReducer,
+const createStore = createStoreFactory({
   group: groupReducer,
+  userHome: userHomeReducer,
   landscape: landscapeReducer,
-  notifications: notificationsReducer,
```
josebui, line 47: Question: Where are the notifications and account reducer defined now?

src/group/membership/components/GroupMembershipCard.js
```diff
@@ -146,11 +146,11 @@ const GroupMembershipCard = props => {
   const { groupSlug } = useGroupContext();
   const { onViewMembers, InfoComponent } = props;
   const { fetching, group } = useSelector(
-    _.getOr({}, `group.memberships.${groupSlug}`)
+    state => state.memberships.memberships[groupSlug] ?? {}
```
josebui, line 149: Question: Is there a reason to remove the lodash use here?

josebui, line 149: Suggestion: Should we change the state structure to: `state.memberships.list` or maybe even `state.memberships.dict` since it is a dictionary? `state.memberships.memberships` seems confusing to me

src/terrasoApi/shared/memberships/membershipsSlice.ts
```diff
@@ -0,0 +1,255 @@
...
+
+type MembershipGroup = {
+  fetching?: boolean;
+  joining?: boolean;
+  message?: Message;
+  group?: MembershipList;
+};
+
+const initialState = {
+  memberships: {} as Record<string, MembershipGroup | undefined>,
```
josebui, line 54: Question does this means that memberships can be undefined?

package.json
```diff
@@ -126,7 +126,7 @@
     "extensions": {
       "codegen": {
         "generates": {
-          "src/terrasoApi/gql/": {
+          "src/terrasoApi/shared/graphqlSchema/": {
```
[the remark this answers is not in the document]
shrouxm: yes, as long as there are fragments and queries defined in the web client codegen will need to happen in the web client. i'm not entirely sure what it will look like as i haven't set that up before, but this config will definitely need to change a bit when that happens

src/terrasoApi/shared/memberships/membershipsFragments.ts
shrouxm: it would be needed for non-group code, i.e. when conceptually thinking of `GroupNode` as `MembershipList`. i should update the fragment names to reflect this. the example right now is `defaultGroupWithMembersSample` in `landscapeFragments` needs to manually remember to add `id` and `slug`, and same with project queries eventually

src/terrasoApi/store.ts
shrouxm: `terrasoApi/shared/shared/store.ts`

src/group/membership/components/GroupMembershipCard.js
shrouxm: i generally have been leaning away from lodash because it breaks jump-to-definition/references & can't be typed properly. with typescript, you can define a `useSelector` that's aware of your state type, so by rewriting this my IDE knows what type group is, which made it easier for me to follow the flow in this function and make sure everything was still working.

shrouxm: i think that might be a good idea, maybe `state.memberships.lists` since it is a dict of `MembershipLists`

josebui: That makes sense to me

josebui: That sounds good to me

david-code (approved):
Looks good Garo. I don't have much to add in terms of suggestions, but it was useful to read the code and see what's going on.

src/group/groupService.ts
```diff
@@ -132,14 +119,18 @@ export const fetchGroups = () => {
...
+          ),
           membersInfo: extractMembersInfo(edge.node),
         }))
       )
       // eslint-disable-next-line lodash-fp/no-extraneous-function-wrapping
-      .then(groups =>
-        _.orderBy([group => group.name.toLowerCase()], [], groups)
-      )
+      .then(groups => {
+        dispatch(setMemberships(getMemberships(groups)));
```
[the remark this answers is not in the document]
shrouxm: it's because i've separated the Terraso Group & membership logic into different services & slices since we aren't moving the Terraso Group logic into the shared code. before, the group slice would be able to update the group info and membership info within the same slice in the same action resolver. but slices must communicate with each other by dispatch. this is already how the landscape slice was working, it just updates group slice to match. 

while checking on the answer to this question i did notice that the landscape slice does this extra dispatch logic in the slice code instead of the service code, so i may update the group code to match when i rebase for consistency (i think it does make sense to keep dispatch code out of the service code).

you're right the eslint comment doesn't make sense. we are using eslint afaik, i think it runs on each PR to the repo

shrouxm:
@paulschreiber i need an approve from you to merge since you requested changes

paulschreiber (approved):

PR #976, opened 2023-05-31
fix: revert assumption that landscape default groups are non-null in client

shrouxm:
## Description
Previous changes assumed that the groups passed into groupUtils would always be non-null. This is currently not the case, so this change undoes that assumption.

### Checklist
- [x] New tests added

### Verification steps
Should load landscapes page ok in environments where landscapes have no default group in DB.

### DIFF
--- a/src/landscape/landscapeService.test.ts
+++ b/src/landscape/landscapeService.test.ts
@@ -1,0 +1,57 @@
+/*
+ * Copyright © 2023 Technology Matters
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


src/landscape/landscapeService.test.ts
```diff
@@ -0,0 +1,57 @@
...
+              description: 'Landscape description',
+              website: 'https://www.landscape.org',
+            },
+          },
+        ],
+      },
+    })
+  );
+  const landscape = await landscapeService.fetchLandscapeToView('');
+  expect(landscape).toStrictEqual({
```
josebui, line 41: Question: What is this test checking? is it that the structure of the landscape data has all the fields defined even if they don't come from the API?

josebui (approved):
Looks good, just one question about the test

src/landscape/landscapeService.test.ts
shrouxm: yes, and that the code doesn't crash if the fields are missing from the API (this test would've crashed the code before the fix).

PR #1003, opened 2023-06-14
fix: increase default group home card margin right

jodiezhuu:
## Description
Increase right margin on groups home card

Screen Shot 2023-06-14 at 10 51 38 AM


src/group/components/GroupDefaultHomeCard.js
```diff
@@ -39,7 +39,7 @@ const GroupDefaultHomeCard = () => {
     >
       <Box display="flex" alignItems="center">
         <Trans i18nKey="group.default_content">
-          <Typography variant="body1">
+          <Typography variant="body1" marginLeft="8px">
```
josebui, line 42: Issue: We should use the spacing from the theme, not sure if it is also 8 px but it should be referenced from the spacing definition from MUI
```suggestion
          <Typography variant="body1" sx={{ ml: 1 }}>
```

josebui (changes requested):

josebui (approved):

PR #1013, opened 2023-06-19
fix: minor text label typo/inconsistencies

shrouxm:
## Description
Just a bit of tidying in the code. I can't actually find the word "Connect" anywhere on the site so it might be unused? But would require digging into the group UI code a bit more so just submitting this quick PR because it is faster than opening an issue.


paulschreiber (approved):

shrouxm:
ah, i see in the failing test the text still said "Connect" in the ARIA label, so this does actually fix something :tada:

PR #1063, opened 2023-07-11
chore: update prettier to version 3.0.0

shrouxm:
updates prettier to version 3.0.0


paulschreiber (changes requested):
`stylelint` is failing. Likely needs a similar change.

shrouxm:
@paulschreiber fixed

PR #1064, opened 2023-07-12
fix: Ignore config bounds in annotate step

### LINKED ISSUE #1033: Bug - Location of demo point is lost in step 4 once step 5 sets viewport away from preview point.
The location of location of a preview point is lost in step 4 once the user moves the viewport away in step 5. 

### Steps to replicate 
1. Use [Modis file](https://firms.modaps.eosdis.nasa.gov/data/active_fire/modis-c6.1/csv/MODIS_C6_1_Global_7d.csv)
2. Progress to step 5
3. Zoom to a part of the map not displayed in step 4
4. return from step 4
5. see issue

### expected behavior
The preview point should be visible
image

### actual behavior
The preview point is not visible because the window has moved away from the preview point
image

josebui:
## Description

 Ignore config bounds in annotate step

### Checklist
- [ ] New tests added

### Related Issues
Fixes #1033


src/sharedData/visualization/components/Visualization.js
```diff
@@ -352,6 +303,48 @@ const MapboxLayer = props => {
...
+        return;
+      }
+      const calculatedBbox = bbox(map.getSource('visualization')._data);
+      return new mapboxgl.LngLatBounds(
+        [calculatedBbox[0], calculatedBbox[1]],
+        [calculatedBbox[2], calculatedBbox[3]]
+      );
+    })();
+
+    const bounds = useConfigBounds ? configBounds : geoJsonBounds;
```
paulschreiber, line 337: question: what does this do? when do you use config bounds vs JSON bounds?

josebui: @paulschreiber we have 2 type of bounds:
1. GeoJSON bounds: These are the bounds calculated from the geo json data
2. Config bounds: These are the bounds explicitly set by the user when they move the viewport

For some steps we don't need to use the config boungs, just the geojson bounds like the annotate step where just one point is displayed

paulschreiber: Should we call it viewportBounds instead of configBounds?

josebui: @paulschreiber for me it makes more sense configBounds, because ideally with our model the bounds could be defined in different ways, and the idea for this condition is to use or not use the bounds defined in the visualization configuration, I think how the user defined those bounds is relevant here

paulschreiber: Maybe visualizationConfigBounds then? ConfigBounds sounded like it was based on the application configuration.

paulschreiber (approved):

PR #1136, opened 2023-08-28
fix: Invalid map bounds

josebui:
## Description

- Added bounds validation in Map component
- Add condition to only add bbox to geojson value if there are features defined and bounding box is defined

### Checklist
- [ ] New tests added

### Related Issues
Fixes #....

### DIFF
--- a/src/gis/components/Map.js
+++ b/src/gis/components/Map.js
@@ -243,6 +243,20 @@ const Map = props => {
   const [bounds] = useState(initialBounds);
 
   useEffect(() => {
+    const isValidBounds = (function () {
+      if (!bounds) {
+        return false;
+      }
+      const [swLng, swLat, neLng, neLat] = bounds;
+      const validLat = value => value >= -90 && value <= 90;
+      const validLng = value => value >= -180 && value <= 180;
+      return (
+        validLat(swLat) && validLat(neLat) && validLng(swLng) && validLng(neLng)
+      );
+    })();
+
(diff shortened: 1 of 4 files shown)


src/gis/components/Map.js
```diff
@@ -22,6 +22,7 @@ import { Box } from '@mui/material';
 
 import { withWrapper } from 'react-hoc';
 
+import { areValidBounds } from 'gis/gisUtils';
```
paulschreiber, line 25:
```suggestion
import { isValidBounds } from 'gis/gisUtils';
```

src/gis/components/Map.js
```diff
@@ -243,6 +244,8 @@ const Map = props => {
   const [bounds] = useState(initialBounds);
 
   useEffect(() => {
+    const validBounds = areValidBounds(bounds);
```
paulschreiber, line 247:
```suggestion
    const validBounds = isValidBounds(bounds);
```

src/gis/gisUtils.js
```diff
@@ -76,3 +76,16 @@ export const parseGeoJson = areaPolygon => {
 };
 
 export const isValidGeoJson = areaPolygon => !!parseGeoJson(areaPolygon);
+
+export const areValidBounds = bounds => {
```
paulschreiber, line 80:
```suggestion
export const isValidBounds = bounds => {
```

paulschreiber (changes requested):

PR #1142, opened 2023-08-29
fix: Removed feature to allow async permissions resolution

josebui:
## Description

When permissions were implemented there was an assumption that permissions resolution could be a async process, but it is not and it is making the logic to handle loding them unnecesary so I will remove them until the need for something like this arises


josebui:
This is not needed right now

PR #1189, opened 2023-09-22
feat: Track new story source

### LINKED ISSUE #1184: Plasubile update: note whether starting a story map happens from home screen or story maps page
We'd like to know, where are story maps being initiated from?

Add property to storymap.start:
- source:
  - home
  - story maps page

josebui:
## Description

- Added property to track new story map source

### Checklist
- [ ] New tests added

### Related Issues
Fixes #1184

### DIFF
--- a/src/storyMap/components/StoryMapsCard.js
+++ b/src/storyMap/components/StoryMapsCard.js
@@ -240,7 +240,8 @@ const StoryMapsCard = ({ title, storyMaps, showCreate = true }) => {
     () =>
       showCreate && {
         label: t('storyMap.home_create'),
-        to: 'tools/story-maps/new?source=home_page',
+        to: 'tools/story-maps/new',
+        pathState: { source: 'home_page' },
       },
     [t, showCreate]
   );
--- a/src/storyMap/components/StoryMapsCard.js
+++ b/src/storyMap/components/StoryMapsCard.js
@@ -240,7 +240,7 @@ const StoryMapsCard = ({ title, storyMaps, showCreate = true }) => {
     () =>
       showCreate && {
         label: t('storyMap.home_create'),
-        to: 'tools/story-maps/new',
+        to: 'tools/story-maps/new?utm_source=home_page',
       },
     [t, showCreate]
   );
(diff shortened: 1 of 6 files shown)


src/storyMap/components/StoryMapsCard.js
```diff
@@ -232,7 +232,7 @@ const StoryMapsCard = ({ title, storyMaps, showCreate = true }) => {
     () =>
       showCreate && {
         label: t('storyMap.home_create'),
-        to: 'tools/story-maps/new',
+        to: 'tools/story-maps/new?source=home_page',
```
paulschreiber, line 235: suggestion: use `utm_source`, which is a standard.

src/storyMap/components/StoryMapsHomeCardDefault.js
```diff
@@ -31,7 +31,7 @@ const StoryMapsHomeCardDefault = () => {
       titleId="story-maps-default-title"
       action={{
         label: t('storyMap.home_create'),
-        to: 'tools/story-maps/new',
+        to: 'tools/story-maps/new?source=home_page',
```
paulschreiber, line 34: suggestion: use `utm_source`, which is a standard.

src/storyMap/components/StoryMapsToolHome.js
```diff
@@ -105,7 +105,7 @@ const StoryMapsToolsHome = () => {
               <Button
                 variant="contained"
                 component={Link}
-                to="/tools/story-maps/new"
+                to="/tools/story-maps/new?source=story_maps_page"
```
paulschreiber, line 108: suggestion: use `utm_source`, which is a standard.

paulschreiber (changes requested):

src/storyMap/components/StoryMapsCard.js
josebui: @paulschreiber I tried that and plausible didn't tack it automatically, so it seems wrong to use it and then add the value to a custom prop

paulschreiber: See https://plausible.io/blog/utm-tracking-tags

josebui: @paulschreiber Reading that page, and doing some research it seems that utm_source is used for external traffic to clarify from where it is comming from, so I changed it to avoid using URL parameters since our app should be able to track internally what happened, let me know what do you think

paulschreiber (approved):

PR #1200, opened 2023-10-02
fix: Added error message when user can't approve membership

### LINKED ISSUE #1196: Bug: user invite to story map failed when user logged in with the wrong account.
User who is invited to edit a story map can't because they are logged in with the wrong account. 

## Exchange
**User**
> Thank you very much for taking my editing suggestions into account! I have "invited" some colleagues to edit my draft storymap (Alvelal), but they tell me they can't open it, it tells them that the storymap is not found.
> I have done what you told me, in "Invite" and there I have put their emails... I don't know if I'm doing something wrong, but they can't access it.

**Derek**
> It looks like you invited two emails: <email1>, <email2> But it doesn't look like we have any users who have accepted those invites. [Person name] will need to accept an invitation she received and create an account in Terraso in order to edit the story map. Could she try to create a Terraso account and tell me if that works?

(Issue gets resolved without Derek knowing what happens - follows up)

**Derek**
>  Our theory is that [Person name] received an email invite at one email address, but ended up creating a Terraso account with another email address. So when she tried to access the story map with an account that wasn't the same one you invited, she ran into this bug.

**User** 
> Yes, that's what happened. [Person name] had previously registered with her gmail account and I invited her with her corporate account alvelal... and that's why she couldn't. And the same with [person name]. But I sent them an invitation to their gmail accounts and both have been able to log in.

Process to recreate:
1. Have terraso member create story map and invites someone
2. Non-terraso member receives invite email 
3. Non-terraso member logs in with a _different_ email address than the one was invited
4. Non-terraso member can't see the new story map to which they have been invited 

### Expected outcome:
Force the user to be logged in before accepting the invitation, that way we could have all the data to identify if the user that is accepting the invite is different from the expected user in the email

### Actual outcome
USer logs in to the wrong account, doesn't see story map in their personal ids. invitee asks for a direct link and, upon receiving, sees a blank page ('story map not found')

josebui:
## Description

- Added error message when user can't approve membership

### Checklist
- [x] New tests added

### Related Issues
Fixes #1196 

### Dependencies
- https://github.com/techmatters/terraso-client-shared/pull/114
- https://github.com/techmatters/terraso-backend/pull/858

### DIFF
--- a/src/account/components/AccountLogin.js
+++ b/src/account/components/AccountLogin.js
@@ -15,6 +15,7 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 import React, { useEffect } from 'react';
+import queryString from 'query-string';
 import { Trans, useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 import { useNavigate, useSearchParams } from 'react-router-dom';
@@ -41,9 +42,23 @@ const MicrosoftIcon = props => {
 };
 
 const appendReferrer = (url, referrer) => {
-  return referrer
-    ? `${url}&state=account?referrerBase64=${btoa(referrer)}`
-    : url;
+  if (!referrer) {
+    return url;
+  }
+  const parsedUrl = queryString.parseUrl(url);
+  const redirectUrl = queryString.stringifyUrl({
+    url: 'account',
+    query: {
+      referrerBase64: btoa(referrer),
+    },
+  });
+  return queryString.stringifyUrl({
+    ...parsedUrl,
+    query: {
+      ...parsedUrl.query,
+      state: redirectUrl,
+    },
+  });
 };
 
 const AccountForm = () => {
--- a/src/account/components/AccountLogin.js
+++ b/src/account/components/AccountLogin.js
@@ -42,7 +42,7 @@ const MicrosoftIcon = props => {
 
 const appendReferrer = (url, referrer) => {
   return referrer
-    ? `${url}&state=account?referrerBase64=${btoa(referrer)}`
+    ? `${url}&state=/account?referrerBase64=${btoa(referrer)}`
     : url;
 };
 
--- a/src/account/components/AccountLogin.js
+++ b/src/account/components/AccountLogin.js
@@ -17,7 +17,7 @@
 import React from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
-import { useSearchParams } from 'react-router-dom';
+import { Navigate, useSearchParams } from 'react-router-dom';
 import { fetchAuthURLs } from 'terraso-client-shared/account/accountSlice';
 import { useFetchData } from 'terraso-client-shared/store/utils';
 import AppleIcon from '@mui/icons-material/Apple';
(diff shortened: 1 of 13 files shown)


src/account/components/AccountLogin.js
```diff
@@ -44,25 +44,45 @@ const GoogleIcon = props => {
   return <SvgIcon component={GoogleSvg} {...props} />;
 };
 
-const appendReferrer = (url, referrer) =>
-  referrer ? `${url}&state=${referrer}` : url;
+const appendReferrer = (url, referrer) => {
+  return referrer
+    ? `${url}&state=account?referrerBase64=${btoa(referrer)}`
+    : url;
+};
```
paulschreiber, line 51: issue: this URL is malformatted. You can't have two `?`s. Please use a URL parsing function to add the arguments.

paulschreiber: Perhaps https://www.npmjs.com/package/query-string will work.

src/account/components/RequireAuth.js
```diff
@@ -48,7 +48,9 @@ const RequireAuth = ({ children }) => {
 
   const referrer = generateReferrerPath(location);
 
-  const to = referrer ? `/account?referrer=${referrer}` : '/account';
+  const to = referrer
+    ? `/account?referrer=/${encodeURIComponent(referrer)}`
+    : '/account';
```
paulschreiber, line 53: suggestion: use the same URL parsing too here.

src/localization/locales/en-US.json
```diff
@@ -1014,7 +1014,8 @@
         "addMemberships.unique": "One or more users are already collaborators of this story map.",
         "approved_membership": "You have been added to “{{storyMapTitle}}”",
         "approve_invite_success": "You have been added to “{{storyMapTitle}}”",
-        "approve_invite_error": "Failed to accept Story Map invite",
+        "approveMembership.update_not_allowed": "Failed to accept Story Map invite",
```
paulschreiber, line 1017:
```suggestion
        "approveMembership.update_not_allowed": "Unable to accept Story Map invitation",
```

src/localization/locales/en-US.json
```diff
@@ -1014,7 +1014,8 @@
...
+        "approveMembership.update_not_allowed_permissions_validation": "You do not have permission to accept this Story Map invite",
```
paulschreiber, line 1018:
```suggestion
        "approveMembership.update_not_allowed_permissions_validation": "You can’t accept this Story Map invitation. You must be logged in using the email address shown in the invitation.",
```

paulschreiber (changes requested):

src/localization/locales/en-US.json
```diff
@@ -1014,7 +1014,8 @@
         "addMemberships.unique": "One or more users are already collaborators of this story map.",
         "approved_membership": "You have been added to “{{storyMapTitle}}”",
         "approve_invite_success": "You have been added to “{{storyMapTitle}}”",
-        "approve_invite_error": "Failed to accept Story Map invite",
+        "approveMembership.update_not_allowed": "Unable to accept Story Map invitation",
+        "approveMembership.update_not_allowed_permissions_validation": "You can't accept this Story Map invitation. You must be logged in using the email address shown in the invitation.",
```
paulschreiber, line 1018:
```suggestion
        "approveMembership.update_not_allowed_permissions_validation": "You can't accept the invitation for the Story Map “{{storyMapTitle}}.” You must be logged in using the email address shown in the invitation.",
```

paulschreiber: also update the code that calls `t()` to pass in the title parameter.

paulschreiber (changes requested):

paulschreiber (approved):

PR #1573, opened 2024-02-27
fix: add fallback page title when first name is missing

paulschreiber:
## Description
Add fallback page title to Terraso home page when first name is missing.


tm-ruxandra (approved):
nit: might be helpful to add a test for displaying the user's name on screen in both cases (seems like there's some precedent for testing content in `Home.test.js`). otherwise lgtm

josebui (approved):

PR #1579, opened 2024-02-28
fix: check empty bounds when fitting map to landscape list

tm-ruxandra:
## Description
Opening the landscape list without any existing landscapes caused an error when the map fit to the bounds of all available landscapes. This is because a newly-initialized `LngLatBounds` object is empty and has no defined latitude or longitude. Fix the issue by checking that the bounds is non-empty before passing it to the `fitBounds` method.

### Checklist
- [x] Verified on mobile
- [x] Verified on desktop

### Verification steps
Open the landscapes view for an account with no current landscapes. Verify that the page opens successfully without displaying an error.


paulschreiber (approved):

paulschreiber (changes requested):
Please fix the failed tests.

paulschreiber (approved):

PR #1589, opened 2024-02-29
build: update @mapbox/mapbox-gl-geocoder@5.0.2 mapbox-gl@3.1.2

paulschreiber:
## Description
update @mapbox/mapbox-gl-geocoder@5.0.2 mapbox-gl@3.1.2.


josebui (approved):
The code changes look good, I just want to restate that mapbox functionality is not covered by the tests, so all functionality should be tested manually, so it should also by tested by product and UX (IMO)

Issue #1400, opened 2024-01-09
While making a data map, selecting "publish" creates a map but does not advance the user to published page.


DerekCaelin:
## Description
While making a data map, selecting "publish" creates a map but does not advance the user to published page. The user receives a notification that the map is created but remains on the preview page. 

## Steps To Reproduce
1. Upload a map file to Terraso, eg [THIS ONE](https://drive.google.com/file/d/196geXcr86UuuvwyyHYwKpewMZLiYyPZk/view?usp=sharing)
2. Make a map of the file
3. Progress to step 4
4. Select "Publish"
5. See issue

## Expected behavior
Data map is created, and the user advances to the published page hosting the data map.

## Actual behavior
Data map is created, but user remains on the "Preview" page of the map making experience. User must navigate via breadcrumbs or back button to leave the data map creation process. 
<img width="1067" alt="image" src="https://github.com/techmatters/terraso-web-client/assets/10799815/d007e507-9c01-48a6-bb42-4e41df1bce8f">

PR #1633, opened 2024-03-14
fix: make nav blocker tool fully disable on form save

tm-ruxandra:
## Description

The navigation blocker as implemented ran into an issue with saving on the map-creation workflow. The form was supposed to navigate to the user's newly-created map on save, but because of how react processes events and state changes, the navigation blocker was still active for the event cycle when the completion callback initiated the new navigation target. This introduces a new capability to the blocker which allows client code to disable the blocking behavior immediately, which allows the navigation to proceed as intended. Also does some minor name refactoring for clarity.

### Checklist
- [x] New tests added

### Related Issues
Fixes [1400](https://github.com/techmatters/terraso-web-client/issues/1400)

### Verification steps

- Create a new landscape, upload a file, and begin creating a map from it ([as described in original report](https://github.com/techmatters/terraso-web-client/issues/1400))
- Verify that clicking to navigate away somewhere else still brings up the blocker
- Verify that saving the map navigates to the newly created map

- Verify that story map creation workflow is unchanged (it used the same navigation blocker hook)

### DIFF
--- a/src/navigation/navigationContext.js
+++ b/src/navigation/navigationContext.js
@@ -20,6 +20,8 @@ import {
   useNavigate,
 } from 'react-router-dom';
 
+let originalPush;
+
 export const useNavigationBlocker = (when, message) => {
   const navigate = useNavigate();
   const { navigator } = useContext(NavigationContext);
@@ -31,8 +33,8 @@ export const useNavigationBlocker = (when, message) => {
    * so that we can be sure we're restoring the original when the blocker is
    * disabled or otherwised cleaned up.
    */
-  if (!navigator.originalPush) {
-    navigator.originalPush = navigator.push;
+  if (!originalPush) {
+    originalPush = navigator.push;
   }
 
   const proceed = useCallback(() => {
@@ -58,7 +60,7 @@ export const useNavigationBlocker = (when, message) => {
 
   const disable = useCallback(() => {
     window.removeEventListener('beforeunload', preventNavigation);
-    navigator.push = navigator.originalPush;
+    navigator.push = originalPush;
   }, [preventNavigation, navigator]);
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


src/navigation/navigationContext.js
```diff
@@ -26,34 +26,52 @@ export const useNavigationBlocker = (when, message) => {
   const [isBlocked, setIsBlocked] = useState(false);
   const [blockedArgs, setBlockedArgs] = useState();
 
-  const unblock = useCallback(() => {
+  /*
+   * we cache a reference to the original push method if we don't have one yet,
```
paulschreiber, line 30:
```suggestion
   * We store a reference to the original push method (if we don't have one yet),
```

paulschreiber (approved):

src/navigation/navigationContext.js
```diff
@@ -26,34 +26,52 @@ export const useNavigationBlocker = (when, message) => {
...
+   * so that we can be sure we're restoring the original when the blocker is
+   * disabled or otherwised cleaned up.
+   */
+  if (!navigator.originalPush) {
+    navigator.originalPush = navigator.push;
```
josebui, line 35: Suggestion: Since `originalPush` is something only relevant within this code, I think it is better if we don't create a new property inside `navigator`, since it is a definition that we don't own, I think we could create a separated originalPush variable, or maybe use useRef or useState if that is needed, but it seems that a variable might work

tm-ruxandra: That's a fair point - this seemed the most straightforward way but I could try to find another. The simple approach of just using a variable in the method body here doesn't work because we'll lose the reference to the original push method after the render runs a second time, which was why I did this (since we were already doing things to the `navigator` object's members).

tm-ruxandra: I tried a few things and ended up settling in a variable for this module. Let me know if you'd like to discuss this further before merging.

josebui (approved):

Issue #1551, opened 2024-02-20
In Terraso shared file module, display extensions for files


DerekCaelin:
We designed an experience for displaying files and extensions for the experience of making maps with data - the experience should extend to the main file/link sharing module.

## Current Implementation
<img width="967" alt="image" src="https://github.com/techmatters/terraso-web-client/assets/92342777/6fd27193-9c94-4730-a2a1-45e2ea17ec56">

## Desired Implementation
<img width="967" alt="image" src="https://github.com/techmatters/terraso-web-client/assets/92342777/22a8a08e-be03-419e-a9b5-6b896ffd71b5">

Hanal-code:
**[MOCKUP] Tabular data layout change between large viewport and tiny viewport**
Note how the file name column(Column 1), metadata column(Column 2) and action column(Column 3) shift
<img width="1561" alt="image" src="https://github.com/techmatters/terraso-web-client/assets/92342777/354dbbe2-ef2c-497b-98ef-ab5afa1b16d1">

PR #1660, opened 2024-03-20
feat: Add file extension type column to shared data card for files

knipec:
## Description
In the "Shared Files and Links" section, shared files now display a column with the file extension type.
This addresses GitHub Issue #1551.

### Checklist
- [x] New tests added
- [x] Verified on mobile
- [x] Verified on desktop

### Related Issues
Will file some other issues found in the process of looking at this, but not directly related. One on grid layout particulars, one on localization, and one on edit description save button.

### Verification steps
1. Go to a landscape and scroll down to the "Shared Files and Links" card
2. Add files if there aren't any already
3. Confirm there is a column for the file extension type, that it displays the correct file extension, that it looks ok on different size screens

### Aside
- The fileSize and resourceType props are only relevant for SharedDataEntryFile, but the concepts exist in SharedDataEntryBase (previously, there was just one prop that was also only relevant for SharedDataEntryFile, but was more generally named 'info')

### DIFF
--- a/src/landscape/components/LandscapeView.test.js
+++ b/src/landscape/components/LandscapeView.test.js
@@ -122,6 +122,7 @@ const baseViewTest = async (
             description: `Description ${index}`,
             size: 3456,
             entryType: 'FILE',
+            resourceType: 'txt',
             visualizations: { edges: [] },
           },
         },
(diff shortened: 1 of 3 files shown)


paulschreiber (changes requested):

paulschreiber:
@knipec I recommend transforming the text to uppercase in the view layer. This also avoids the null dereference causing test failures.

paulschreiber (approved):

PR #1669, opened 2024-03-25
fix: shared files layout improvements

### LINKED ISSUE #1659: Vertically align the file type icon with top  in Shared Files & Links module
## Description
When a file name wraps to multiple lines the icon shows up in the center of the cell vertically. It looks inconsistent from the single line item.  It also makes the tabular list less scannable.

## Steps To Reproduce

1. Go to Group or Landscape page: [[insert URL here](https://app.staging.terraso.net/groups/academic-research-test)]
2. Scroll down to 'Shared files and Links' module
3. Note  a long file/link title gets wrapped and the icon gets aligned center.

## Expected behavior
image

## Actual behavior
image

paulschreiber:
## Description
* vertically align file type icon on mobile
* add space at top of full-screen confirmation dialog
* prevent share icon from overlapping file name on mobile

### Checklist
- [X] Verified on mobile
- [X] Verified on desktop

### Related Issues
Fixes #1659.

### DIFF
--- a/src/sharedData/components/SharedDataEntryBase.js
+++ b/src/sharedData/components/SharedDataEntryBase.js
@@ -20,6 +20,7 @@ import { useTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
 import DeleteIcon from '@mui/icons-material/Delete';
 import { Grid, ListItem, Stack, Typography } from '@mui/material';
+import useMediaQuery from '@mui/material/useMediaQuery';
 
 import { daysSince } from 'timeUtils';
 
@@ -69,6 +70,7 @@ const SharedDataEntryBase = props => {
   );
   const dispatch = useDispatch();
   const { trackEvent } = useAnalytics();
+  const isSmall = useMediaQuery(theme.breakpoints.down('sm'));
 
   const onConfirm = useCallback(() => {
     dispatch(deleteSharedData({ dataEntry, sharedResource })).then(data => {
@@ -182,7 +184,10 @@ const SharedDataEntryBase = props => {
           order={{ xs: 2, md: 2 }}
           component={StackRow}
         >
-          <EntryTypeIcon resourceType={dataEntry.resourceType} />
+          <EntryTypeIcon
+            resourceType={dataEntry.resourceType}
+            style={isSmall ? { marginTop: '-1em', paddingRight: '5px' } : {}}
+          />
           <Restricted
             permission="sharedData.edit"
             resource={permissionsResource}
--- a/src/sharedData/components/SharedDataEntryBase.js
+++ b/src/sharedData/components/SharedDataEntryBase.js
@@ -20,7 +20,6 @@ import { useTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
 import DeleteIcon from '@mui/icons-material/Delete';
 import { Grid, ListItem, Stack, Typography } from '@mui/material';
-import useMediaQuery from '@mui/material/useMediaQuery';
 
 import { daysSince } from 'timeUtils';
 
@@ -70,7 +69,6 @@ const SharedDataEntryBase = props => {
   );
   const dispatch = useDispatch();
   const { trackEvent } = useAnalytics();
-  const isSmall = useMediaQuery(theme.breakpoints.down('sm'));
 
   const onConfirm = useCallback(() => {
     dispatch(deleteSharedData({ dataEntry, sharedResource })).then(data => {
@@ -183,10 +181,20 @@ const SharedDataEntryBase = props => {
           md={5}
           order={{ xs: 2, md: 2 }}
           component={StackRow}
+          sx={{
+            alignItems: {
+              xs: 'flex-start',
+              md: 'inherit',
+            },
+          }}
         >
           <EntryTypeIcon
             resourceType={dataEntry.resourceType}
-            style={isSmall ? { marginTop: '-1em', paddingRight: '5px' } : {}}
+            styleProps={{
+              marginTop: { xs: '1em', md: 'inherit' },
+              marginLeft: { xs: '-5px', md: 'inherit' },
+              paddingRight: { xs: '5px', md: 'inherit' },
+            }}
           />
           <Restricted
             permission="sharedData.edit"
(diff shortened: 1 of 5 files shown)


src/sharedData/components/SharedDataEntryBase.js
```diff
@@ -169,16 +171,23 @@ const SharedDataEntryBase = props => {
...
-          xs={isEditingName ? 12 : 8}
+          xs={isEditingName ? 12 : 7}
           md={5}
           order={{ xs: 2, md: 2 }}
           component={StackRow}
         >
-          <EntryTypeIcon resourceType={dataEntry.resourceType} />
+          <EntryTypeIcon
+            resourceType={dataEntry.resourceType}
+            style={isSmall ? { marginTop: '-1em', paddingRight: '5px' } : {}}
```
knipec, line 189: I don't understand this approach -- I tried these changes and it doesn't seem to line up the top of the icon with the top of the title's first line of text. (The title can also be a variable number of lines long.)

Would we instead want `alignItems: 'flex-start'` somewhere, maybe with some padding?

src/sharedData/components/SharedDataEntryBase.js
```diff
@@ -233,7 +242,7 @@ const SharedDataEntryBase = props => {
 
         <Grid
           item
-          xs={4}
+          xs={5}
```
knipec, line 245: The description no longer lines up with the start of the buttons -- might be worth confirming with Hana if this is ok

knipec (changes requested):
Could you explain why your approach for the icon vertical alignment works? I might be missing something.

src/sharedData/components/SharedDataEntryBase.js
```diff
@@ -69,6 +70,7 @@ const SharedDataEntryBase = props => {
   );
   const dispatch = useDispatch();
   const { trackEvent } = useAnalytics();
+  const isSmall = useMediaQuery(theme.breakpoints.down('sm'));
```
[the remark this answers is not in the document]
paulschreiber: `theme.breakpoints.down('xs')` would find things smaller than xs, meaning the styles I want to apply to xs won't apply.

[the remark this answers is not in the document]
knipec: Ah yes, my bad -- should have been "Why 'sm' instead of 'md'?" (Though may not be relevant depending on the rest of the comments)

src/sharedData/components/SharedDataEntryBase.js
paulschreiber: Good idea. I've used flex-start.

src/sharedData/components/SharedFileIcon.js
```diff
@@ -39,13 +40,19 @@ const ICON_FILES = {
   png: 'png.png',
 };
 
-const SharedFileIcon = ({ resourceType }) => {
+const SharedFileIcon = props => {
+  const { resourceType, styleProps } = props;
+  console.log(styleProps);
```
knipec, line 45: Do you want the console.log here?

src/sharedData/components/SharedFileIcon.js
```diff
@@ -39,13 +40,19 @@ const ICON_FILES = {
   png: 'png.png',
 };
 
-const SharedFileIcon = ({ resourceType }) => {
+const SharedFileIcon = props => {
+  const { resourceType, styleProps } = props;
```
knipec, line 44: Oh, whatever we do to the SharedFileIcon we should probably also do to the Link icon

src/sharedData/components/SharedDataEntryBase.js
```diff
@@ -169,16 +169,33 @@ const SharedDataEntryBase = props => {
...
+              xs: 'flex-start',
+              md: 'inherit',
+            },
+          }}
         >
-          <EntryTypeIcon resourceType={dataEntry.resourceType} />
+          <EntryTypeIcon
+            resourceType={dataEntry.resourceType}
+            styleProps={{
+              marginTop: { xs: '1em', md: 'inherit' },
```
knipec, line 194: This looks to me like a bit more space than we'd want at the top but perhaps that is for Hana to say :)

paulschreiber: It lines the icon up with the text.

src/sharedData/components/SharedDataEntryBase.js
```diff
@@ -169,16 +169,33 @@ const SharedDataEntryBase = props => {
...
+            },
+          }}
         >
-          <EntryTypeIcon resourceType={dataEntry.resourceType} />
+          <EntryTypeIcon
+            resourceType={dataEntry.resourceType}
+            styleProps={{
+              marginTop: { xs: '0.8em', md: 'inherit' },
+              marginLeft: { xs: '-5px', md: 'inherit' },
+              paddingRight: { xs: '5px', md: 'inherit' },
```
knipec, line 196: Suggestion/Nitpick: Move this styling to SharedDataEntryFile.js

SharedDataEntryBase.js is used for both shared files and links. 
So if the link icon styling is in SharedDataEntryLink.js because it requires different styling than the shared files icon, it could be more consistent to have this be in the code that is specific to files.

knipec: I'm not suggesting combining them, I'm suggesting moving the styling that is specific to the **FileIcon** to the SharedDateEntry**File**.js file

src/sharedData/components/SharedFileIcon.js
```diff
@@ -39,13 +40,19 @@ const ICON_FILES = {
   png: 'png.png',
 };
 
-const SharedFileIcon = ({ resourceType }) => {
+const SharedFileIcon = props => {
+  const { resourceType, styleProps, fallbackStyleProps } = props;
```
knipec, line 44: nitpick: You can do the destructuring in one line if you want
`const SharedFileIcon = ({ resourceType, styleProps, fallbackStyleProps }) => { 
`
and not need `const { resourceType, styleProps, fallbackStyleProps } = props;`

like how it was before with `({ resourceType })`

knipec (approved):
Thanks for going through this with me!

knipec (approved):

PR #1694, opened 2024-03-28
fix: Display the domain url for links that was previously missing

### LINKED ISSUE #1690: Web URL missing from preview in shared file module
## Description
Shared links used to reveal an extension 

## Steps To Reproduce

1. Go to URL: https://app.terraso.org/landscapes/sacred-sierra
2. Join the landscape
3. Scroll down to the shared file module where there is at least one URL shared

## Expected behavior
We see the preview URL in the area where the file extension is visible
image

## Actual behavior
the preview URL extends over the extension and the size field.
image

knipec:
## Description
Displaying the domain was tied to the `info` prop. The bug was caused when I removed/renamed `info`, such that it then only applied to files, and failed to pass the domain info through for links. 
The `info` concept is now replaced by `InfoComponent` -- whose implementation will differ for files and links, as they will show different info with different layouts.

### Checklist
- [x] New tests added
- [x] Verified on mobile
- [x] Verified on desktop

### Related Issues
Fixes #1690 
(which was caused by #1660)

### Verification steps
See #1690

### DIFF
--- a/src/sharedData/components/SharedDataEntryFile.js
+++ b/src/sharedData/components/SharedDataEntryFile.js
@@ -66,7 +66,7 @@ const InfoComponent = ({ sharedResource }) => {
   const fileSize = filesize(sharedResource.dataEntry.size, { round: 0 });
   const fileType = sharedResource.dataEntry.resourceType;
   return (
-    <Grid container>
+    <>
       <Grid
         item
         xs={3}
@@ -79,7 +79,7 @@ const InfoComponent = ({ sharedResource }) => {
         {fileSize}
       </Grid>
       <Grid item xs={1} display={{ md: 'none' }} />
-    </Grid>
+    </>
   );
 };
 
--- a/src/sharedData/components/SharedDataEntryFile.js
+++ b/src/sharedData/components/SharedDataEntryFile.js
@@ -395,10 +395,6 @@ const SharedDataEntryFile = props => {
       DownloadComponent={DownloadComponent}
       ShareComponent={ShareComponent}
       InfoComponent={InfoComponent}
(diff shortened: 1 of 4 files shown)


src/sharedData/components/SharedDataEntryFile.js
```diff
@@ -62,6 +62,27 @@ const StackRow = props => (
   <Stack direction="row" alignItems="center" spacing={1} {...props} />
 );
 
+const InfoComponent = ({ sharedResource }) => {
+  const fileSize = filesize(sharedResource.dataEntry.size, { round: 0 });
+  const fileType = sharedResource.dataEntry.resourceType;
+  return (
+    <Grid container>
```
paulschreiber, line 69:
```suggestion
    <>
```

src/sharedData/components/SharedDataEntryFile.js
```diff
@@ -62,6 +62,27 @@ const StackRow = props => (
...
+        md={6}
+        sx={{ wordWrap: 'break-word', textTransform: 'uppercase' }}
+      >
+        {fileType}
+      </Grid>
+      <Grid item xs={8} md={6} sx={{ wordWrap: 'break-word' }}>
+        {fileSize}
+      </Grid>
+      <Grid item xs={1} display={{ md: 'none' }} />
+    </Grid>
```
paulschreiber, line 82:
```suggestion
    </>
```

paulschreiber: suggestion: this seems to work fine without an extra container.

paulschreiber (approved):

knipec:
Verified on staging by Derek.

PR #1753, opened 2024-04-09
fix: don't show edit link for pending maps (will result in 404)

paulschreiber:
## Description
Don't show edit link for pending maps (will result in 404).

### DIFF
--- a/src/storyMap/components/StoryMapsCard.js
+++ b/src/storyMap/components/StoryMapsCard.js
@@ -165,16 +165,25 @@ const StoryMapListItem = props => {
           )}
           <CollaborationIndicator storyMap={storyMap} />
         </Stack>
-        <RouterLink
-          id={`story-map-${storyMap.slug}-link`}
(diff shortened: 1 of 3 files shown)


src/storyMap/components/StoryMapsCard.js
```diff
@@ -165,16 +165,25 @@ const StoryMapListItem = props => {
...
-          }
-        >
-          {storyMap.title}
-        </RouterLink>
+        {/*
+          Unpublished drafts you have not accepted get no link, as it would 404.
+          Published and not accepted get a view link.
+          Published and accepted get an edit link.
+        */}
+        {!storyMap.isPublished && isPending ? (
```
josebui, line 173: Suggestion: I think we can improve this name `isPending` from the name and the component name feels like it should be a pending story map, I think this could be `isMembershipPending`

josebui (approved):
Looks good, just one suggestion

Issue #2371, opened 2024-09-10
When the same data layer is used in consecutive chapters, scrolling up causes data layer to disappear


DerekCaelin:
## References
Sample story map: https://app.staging.terraso.net/tools/story-maps/856b720b/data-layers/edit (shared with Jose)
Published story map on production: https://app.terraso.org/tools/story-maps/d500d536/unraveling-the-web-climate-urbanization-and-the-dengue-crisis-in-bangalore

## Steps to reproduce

1. Create story map
2. Add chapter and add a map layer
3. Add chapter and add the **same** map layer
4. Publish the map
5. Scroll to the bottom of the story map
6. Scroll up

## Expected Behavior
Map layers are visible

## Actual behavior
Map layers are invisible

## Additional context
Issue happens only on repeat map layers. Using different map layers means the issue will not occur.

DerekCaelin:
I added 3 chapters with the same map layer to [this story map](https://app.staging.terraso.net/tools/story-maps/856b720b/data-layers) and can successfully scroll up and down without content disappearing. The fix seems to work. @josebui please push to prod.

PR #2419, opened 2024-09-23
fix: Filter layers that used on the current transition

josebui:
## Description

On chapter exit transition filter layers that are used on the current transition

### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/2371


paulschreiber (approved):
Looks good. Do we need to add any tests for this?

josebui:
@paulschreiber I was thinking yesterday about that, it might not be that trivial to add test for this since it is tighly linked to the scrollama library, but let me check if I can do something

paulschreiber (approved):

Issue #2671, opened 2025-04-16
After autosave, the story map "jumps" in the editor


DerekCaelin:
## Description
When the autosave is complete, the storymap "jumps" - a chapter moves up or down, sometimes out of view. 

## Steps To Reproduce

1. Work on a story map (the story map I am on has 20 chapters)
2. make an edit, watch the "Saving" indicator
3. Once the autosave has completed, see issue

## Expected behavior
Story map remains stationary as I work on it.

## Actual behavior
The story map "jumps" - a chapter moves up or down, sometimes out of view

https://github.com/user-attachments/assets/7ab8a072-00d1-45e9-8e15-4d4ae1410028

Console log
`Failed to evaluate expression "["all",["case",["has","reflen"],["<=",["number",["get","reflen"]],6],["has","shield_beta"]],["match",["get","class"],["pedestrian","service"],false,true],["step",["zoom"],["==",["geometry-type"],"Point"],11,[">",["number",["get","len"]],5000],12,[">",["number",["get","len"]],2500],13,[">",["number",["get","len"]],1000],14,true],["step",["pitch"],true,50,["<",["distance-from-center"],2],60,["<",["distance-from-center"],2.5],70,["<",["distance-from-center"],3]]]". The expression ["get","len"] evaluated to null but was expected to be of type number.`

josebui:
@DerekCaelin I deployed a new fix for this to staging, let me know if it looks good now

CourtneyLee333:
Looks good to me!

PR #2726, opened 2025-06-20
fix: Prevent jump in Rich Text Editor

josebui:
## Description

- Added fix to prevent text editor scroll when text editor is visible

### Checklist
- [ ] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/2671

### DIFF
--- a/src/common/components/RichTextEditor/index.js
+++ b/src/common/components/RichTextEditor/index.js
@@ -15,7 +15,13 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-import React, { useCallback, useMemo, useState } from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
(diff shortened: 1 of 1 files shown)


josebui:
@garobrik 

> i'm guessing from context that the story map editor was jumping around due to react re-renders causing elements to re-load media and then have to recompute height?

Yes, it was jamuping beacuse of the height resize while loading or refreshing the media content. I will try to add a comment.

> also, do you know why the test coverage check is failing? i think i've seen that before on web client, and it doesn't seem ideal to have a required CI check which we frequently ignore/override

The coverage is not correctly identifying just the coverage of the changes in the PRs, so it has only be used as reference to check the results. In this case I think it is correct since the effort to add coverage for this is too complex. So, I think it is valuable to have it, I do think it is confusing that it is almost always failing

PR #2744, opened 2025-07-10
feat: modify login to account for multiple domains

garobrik:
## Description
Modifies the login process to account for being able to log in from a different domain, so that we can support preview instances.

When we merge this, I propose setting the staging frontend to auto-deploy from the `main` branch, and getting rid of the `staging` branch. That way we can always have a frontend at a canonical URL which is identical to the prod frontend, but pointed at the staging backend. Once we do that, all PRs opened against the `main` branch will automatically get a preview instance deployed, so we will still be able to test changes, but now in isolation!

### Checklist
- [ ] New tests added
- [ ] English strings reviewed and copyedited
- [ ] Strings are localized
- [ ] Images are compressed (TinyPNG or svgo)
- [ ] Sample environment file updated (when environment variables changed)
- [ ] Verified on mobile
- [ ] Verified on desktop
- [ ] Verified accessibility ([Accessibility - Web applications](https://docs.google.com/document/d/1VwPDyLqw7r_iISMgDZr1FdAmAqqOIIVIxQqsYA-HaHE/edit?usp=sharing))

### Related Issues
Related to techmatters/terraso-backend#1709

### Verification steps
Staging should still work when set to this branch, and preview branches should work when based off this branch.


src/account/components/AccountAuthCallback.tsx
```diff
@@ -0,0 +1,33 @@
+import { useEffect } from 'react';
```
paulschreiber, line 1: issue: add copyright notice

src/account/components/AccountAuthCallback.tsx
```diff
@@ -0,0 +1,33 @@
...
+  rtoken: string;
+  redirectUrl: string;
+};
+const AccountAuthCallback = () => {
+  const navigate = useNavigate();
+  const dispatch = useDispatch();
+  const [searchParams] = useSearchParams();
+  const rawState = searchParams.get('state');
+  useEffect(() => {
+    if (TERRASO_ENV === 'production' || !rawState) {
```
paulschreiber, line 20: issue: render already uses `ENV` for this — would be good not to duplicate this.

paulschreiber (changes requested):

Issue #2605, opened 2025-01-07
Update to React 19


DerekCaelin:
Placeholder: We currently use React 18.? - update react in a testing environment, follow up on broken things, and once ready, update the production app.

PR #2844, opened 2025-10-09
feat: upgrade deps including react 19 redux 2

garobrik:
## Description
There are still a couple more issues to fix, but everything is wired up correctly now!
- [x] There is literally one single test which is still failing.
- [x] The svg import from `client-shared` specifically is not working correctly (google logo doesn't render properly on home screen)
- [x] The depcheck lint is failing

For review: almost all the changed files are in the commit `chore: run refactoring scripts`, which does rote changes like modifying file extensions or updating import formats. All the other commits should be a manageable size to review!

### Checklist

### Related Issues
Related to #2605 and https://github.com/techmatters/terraso-product/issues/563

### Verification steps
Everything should work the same!


josebui (approved):
I think I caught all of the not import related changes and they look good to me

Issue #2862, opened 2025-11-11
User receives unauthenticated error when viewing embeded story maps.


DerekCaelin:
For people who were logged in
automatically logged out
embeds (and views of story maps) don't load.

public views (like embeds) shouldn't expect to send log in information.

ssfrancis:
I just re-tested and could not reproduce.

DerekCaelin:
Closing for now, since the changes Jose made should be on the backend. Will reopen if it occurs again

PR #2867, opened 2025-11-17
fix: Handle signout on optional auth

josebui:
## Description

Handle sign out for Optional Auth

### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/2862

### DIFF
--- a/src/account/accountProfileUtils.js
+++ b/src/account/accountProfileUtils.js
@@ -63,8 +63,8 @@ export const useCompleteProfile = () => {
       .then(isFirstLogin => {
         setIsFirstLogin(isFirstLogin);
       })
-      .catch(() => {
-        setIsFirstLogin(false);
+      .catch(error => {
+        console.error('Failed to check first login status:', error);
       });
   }, []);
 


src/account/accountProfileUtils.js
```diff
@@ -56,12 +56,16 @@ export const useCompleteProfile = () => {
...
-    getIsFirstLogin().then(isFirstLogin => {
-      setIsFirstLogin(isFirstLogin);
-    });
+    getIsFirstLogin()
+      .then(isFirstLogin => {
+        setIsFirstLogin(isFirstLogin);
+      })
+      .catch(() => {
+        setIsFirstLogin(false);
+      });
```
garobrik: question: i'm not understanding what's going on here: trying to follow the code, it seems that since the `useEffect` has no dependencies it will only run once, and since `isFirstLogin` is now default `false`, `setIsFirstLogin(false)` should have no effect? i'm also not sure how this is related to the optional auth fix

src/account/components/OptionalAuth.js
```diff
@@ -37,6 +39,12 @@ const OptionalAuth = ({ children }) => {
     )
   );
 
+  useEffect(() => {
+    if (fetching === false && !user && hasToken) {
+      dispatch(signOut());
```
garobrik: question: in `RequireAuth`, we dispatch `setHasToken(false)` instead of `signOut()` in a similar condition, why is it different?

garobrik (approved):
ok, i think it looks good! (but the auth code generally is still a bit hazy for me). btw this seemed like a difficult bug to repro, were you able to actually recreate the bug conditions in a running backend/frontend (& if so how?) or will we need to deploy to see if it works?

Issue #2813, opened 2025-09-23
Story Maps: Collapsible sidebar to organize user actions


DerekCaelin:
Implement a collapseable sidebar to host featured image, description, and actions currently in the header. 

## References
An extension of https://github.com/techmatters/terraso-web-client/issues/2812 

## Design
[Mural](https://app.mural.co/t/techmatters9109/m/techmatters9109/1752602307859/3b4d4f349913e77236bf602b926d704f159d7dc4?wid=0-1757445352638) - implement Flow 3

<img width="1311" height="284" alt="Image" src="https://github.com/user-attachments/assets/379878a7-9a4d-4f31-8706-1ea2d7974d83" />

## Acceptance Criteria
- [ ] As a user in the story maps editor, I see a sidebar which I can collapse and expand
   - [ ] In the sidebar I have the following affordances
      - [x] Preview image
      - [x] Set featured image
      - [x] Set description
      - [x] See status (published, draft)
      - [x] See "share" button which opens the share experience (link, embed code, social)
      - [x] See contributors 
      - [x] Invite contributors

CourtneyLee333:
@DerekCaelin I've made some [notes in Mural](https://app.mural.co/t/techmatters9109/m/techmatters9109/1752602307859/3b4d4f349913e77236bf602b926d704f159d7dc4?sender=u7b3b2fb95ea53b5320ac3477). Let's discuss, make some decisions, then I can put together a handoff-ready design for Jose.

CourtneyLee333:
@josebui [Here is the Mural design](https://app.mural.co/t/techmatters9109/m/techmatters9109/1752602307859/3b4d4f349913e77236bf602b926d704f159d7dc4?wid=0-1770050533773) that encapsulates our decisions from last week's meeting.

DerekCaelin:
At team meeting, we decided
- remove button on map to un-collapse side menu
- Address bug with resizing screen

<img width="500" src="https://github.com/user-attachments/assets/8f7ff938-6c8a-45e8-b692-5ae6d1978759" />

Please push to production once addressed.

We agreed to reconsider mobile experience.

PR #2941, opened 2026-01-20
feat: Story map right sidebar

josebui:
## Description

- Added Right story map editor sidebar
- Fixed commit checks

### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/2813

### Verification steps

Screenshot 2026-01-20 at 1 33 36 PM
Screenshot 2026-01-20 at 1 33 30 PM


garobrik (approved):
LGTM! i also ran into some issues with the manual JS sizing of the map while implementing pinned map, so there'll be some merge conflicts there. probably i will have to deal with them at this pace 🙃 but i think we're both taking a similar approach (flex sizing, resize observer on the map container) so hopefully shouldn't be too annoying!

PR #2971, opened 2026-03-10
fix: Optional auth landscapes

josebui:
## Description

- Fixed optional auth
- Fixed tests async support

### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/2862

### DIFF
--- a/src/account/components/OptionalAuth.jsx
+++ b/src/account/components/OptionalAuth.jsx
@@ -36,30 +36,34 @@ const OptionalAuth = ({ children }) => {
     user,
   });
 
+  const hasStoredToken = hasToken;
+  const hasResolvedNoUser = !user;
+  const validationFinished = validationAttempted && !validationPending;
+  const accountFetchFinished = fetching === false;
+  const accountFetchPending = !accountFetchFinished;
+
+  const tokenValidationFailed =
+    hasStoredToken &&
+    hasResolvedNoUser &&
+    validationFinished &&
+    accountFetchFinished;
+
+  const awaitingTokenValidationResult =
+    validationPending || !validationAttempted;
+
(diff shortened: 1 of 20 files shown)


src/account/components/OptionalAuth.jsx
```diff
@@ -15,37 +15,55 @@
...
+    accountFetchFinished;
+
+  const awaitingTokenValidationResult =
+    validationPending || !validationAttempted;
+
+  const shouldSignOut = tokenValidationFailed;
+  const shouldShowLoader =
+    hasStoredToken &&
+    hasResolvedNoUser &&
+    (accountFetchPending || awaitingTokenValidationResult);
```
garobrik, line 58: these intermediate results are really helpful for understanding what is happening! is it possible to make `RequireAuth.jsx` read like this as well?

src/permissions/index.jsx
```diff
@@ -48,44 +58,79 @@ export const PermissionsProvider = ({ rules, children }) => {
...
+
+    const result = isAllowedTo(permission, user, resource);
+
+    if (result && typeof result.then === 'function') {
+      return {
+        type: 'async',
+        result: Promise.resolve(result).catch(() => false),
+      };
+    }
+
```
garobrik, line 97: from my reading, it seems that all rules now result in a raw boolean. is that wrong, or is this branch not needed? is it in anticipation of future potential rules that might be async?

garobrik (approved):
i didn't really fully understand the auth flow in the first place and that is still true after reading this PR, but i don't have any concerns about anything i read in the code. 

for my understanding, could you explain why there was an unauthenticated error before and why there won't be now?

PR #2978, opened 2026-03-26
feat: area of interest

### LINKED ISSUE #2934: Area of interest
Prototype and iterate an experience to allow the editor of a story map to to define an area that must be seen regardless of view. 

## Acceptance criteria
- [x] Convert the "edit map" experience to a dialog window
- [x] Set the bounds of the viewport  to define the area of interest
- [x] Programmatically reposition the map based on the dimensions of viewer's window and the position of the chapters

Consider:
- [ ] better to have map customization experience combined with chapter editor experience?
- [ ] what is the best way to set the viewport?

garobrik:
### Remaining work
- [x] change "edit map" to windowed dialog
- [x] issue: swapping chapter alignment on pre-existing story map without editing the map location will cause incorrect area of interest calculations
- [x] issue: map layers aren't getting disabled when chapter is exited anymore
- [x] i haven't checked how this works where longitude wraps around in the pacific :shrug:
- [x] fix tests 
- [x] see about if there's any way to test this ( :grimacing: )

### Changes
i was having a lot of trouble getting this working without a significant code restructuring, since before the map would only move because of scroll events and so the map move was triggered directly in the Scroller component. but now many different events can trigger a map adjustment. while i was making these changes, i was running into issues with other pieces of code which were for features we don't actually expose to the user. 

rather than try to keep those features functional without testing them properly, i've opted to remove them from the codebase, with the idea we can add them back in later if we decide to prioritize them on our roadmap and test them properly at that point. none of the features i removed are on our roadmap currently. here's the list:
- inset map (small minimap in the corner which shows where you are looking on a globe)
- setting a duration for map layer opacity changes
- including a rotation in the animation
- moving the map instantaneously instead of with an animation
- automatically paging through the map on a timer

i also made the following simplifications:
- almost all the logic in Scroller got moved to stateless functions in a new `storyMap/mapUtils.ts` file, which also contains the new GIS code for adjusting map bounds
- there's no more calculating previous and next chapter in order to figure out which layers to disable. instead we just track next chapter, and disable all layers besides the ones in that chapter
- no longer tries to avoid calling mapbox animation by checking the current location. the old code didn't account for zoom, and it was hard to update it to work with bounds. my thinking is that there's no harm in calling it again if you happen to be in the same location, it'll just be a no-op, but let me know if there's any concerns with this.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #2934 

### Verification steps
- Mess around in existing maps on mobile and see that the view alignment is better.
- Edit a story map by changing map location and swapping chapter alignment, resizing browser window, opening/closing sidebar, etc, and see that area of interest remains visible.

### DIFF
--- a/src/storyMap/storyMapTypes.ts
+++ b/src/storyMap/storyMapTypes.ts
@@ -82,7 +82,7 @@ export type StoryMapConfig = {
   chapters: ChapterConfig[];
   titleTransition?: Transition;
   projection?: string;
-  dataLayers: Record<string, MapLayerConfig>;
+  dataLayers?: Record<string, MapLayerConfig>;
 };
 
 export type VisualizationConfigForm = {
--- a/src/storyMap/storyMapTypes.ts
+++ b/src/storyMap/storyMapTypes.ts
@@ -16,7 +16,7 @@
  */
 
 import type { GeoJSON } from 'geojson';
-import { LngLat, LngLatBounds } from 'mapbox-gl';
+import { LngLat } from 'mapbox-gl';
 import { Descendant } from 'slate';
 import type {
   DataEntryNode,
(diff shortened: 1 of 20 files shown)


src/storyMap/storyMapTypes.ts
```diff
@@ -39,35 +39,37 @@ export type MapPosition = {
...
 };
 
+export type LayerConfig = {
+  layer: string;
+  opacity: number;
+};
+
+export type Transition = {
+  location: MapPosition;
+  onChapterEnter?: LayerConfig[];
```
josebui, line 52: Question: Is chapter exit not needed anymore? My suspcision is that it was not funcionaly making sense since maps change just on enter, is that the case?

src/storyMap/storyMapUtils.js
```diff
@@ -15,12 +15,16 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
+/** @import { ChapterConfig, StoryMapConfig, Transition } from 'terraso-web-client/storyMap/storyMapTypes' */
```
josebui, line 18: Question: Is this comment needed?

josebui (approved):
Looks good, just some small questions

src/storyMap/storyMapUtils.js
garobrik: yes, it's a way to import types into javascript files, so that some parts of javascript files can be given TypeScript types without having to convert the whole file or move the code out of the file!

src/storyMap/storyMapTypes.ts
garobrik: there's actually a lot of config parameters we don't use, but i haven't removed them from the code yet because i wanted to think through things before making any actual changes to the underlying data, in case it causes any problems with the preview branch/staging running at the same time or if we need to revert changes.

in general, i'd really like to define the story map config more rigorously and have a way to be more confident when changing its shape going forward.

PR #2985, opened 2026-04-07
feat: Rich Text Editor improvements

josebui:
### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-web-client/issues/3004


src/common/components/RichTextEditor/index.jsx
```diff
@@ -491,67 +769,109 @@ const RichTextEditor = props => {
...
-                  outline: 'none',
-                  '&:focus': focusOutline,
-                }
-              : null
+      <Box
+        ref={containerRef}
+        onFocus={() => setFocused(true)}
+        onBlur={event => {
+          if (event.currentTarget.contains(event.relatedTarget)) {
+            return;
```
garobrik, line 789: what behavior does this new onBlur handle cause?

src/common/components/RichTextEditor/index.jsx
```diff
@@ -97,9 +168,87 @@ import { focusOutline } from 'terraso-web-client/theme';
...
+
+export const setBlockType = (editor, format) => {
+  unwrapLists(editor);
+  Transforms.setNodes(editor, { type: format });
+};
+
+export const toggleBlock = (editor, format) => {
+  const isActive = isBlockActive(editor, format);
+
+  unwrapLists(editor);
```
garobrik, line 227: what does it mean to unwrap or wrap a node and why are we doing it here?

src/common/components/RichTextEditor/index.jsx
```diff
@@ -454,31 +702,61 @@ const RichTextEditor = props => {
...
+
+      onChange(nextValue);
+    },
+    [editor, onChange]
+  );
+
+  const parsedValue = useMemo(() => normalizeValue(value), [value]);
+
+  useEffect(() => {
+    if (areValuesEqual(editor.children, parsedValue)) {
```
garobrik, line 735: does this mean we stringify the JSON contents of the user's input text twice on each keystroke? (idk if this is actually a problem with the size of content our users are inputting, just trying to understand the logic/be aware this is happening in case of future perf issues)

garobrik (commented):
hey jose, i've had time to review the first commit of this PR (rich text editor), there's a few questions but no concerns! i haven't had time to review the commits related to buffering the story map saves yet, since this was a pretty big PR. 

i'm not sure if the rich text editor changes are ok to merge without the buffering, but if they are you're welcome to split the PR and merge just the first commit!

unfortunately i'm out next week, but i will make sure to get to the buffering stuff ASAP when i'm back

src/common/components/RichTextEditor/index.jsx
josebui: This was added to be able to use the heading and body menu of the new toolbar element

src/common/components/RichTextEditor/index.jsx
josebui: From the rich text editro point of view wrap means apply the format and unwrap means unapply the format, in this case apply or unapply the lists formatting. It is defined here because that is the responsability of the rich text editor, to apply or unapply the formatting, I think the others wrapping and unwrapping are also defined here

src/common/components/RichTextEditor/index.jsx
josebui: This was part of the improvement performances attempts, on the complete changes there is a focused conditional the exists early, so while the user is typing this is not triggered. I am going to try to separate the performance optimizations from the richtext editor features, I hope I can sepearte them

garobrik (approved):

PR #2995, opened 2026-04-23
fix: chapter scroll detection

### LINKED ISSUE #2983: Embeded story map doesn't detect chapter transition, chapters semi-transparent
When embedded, story map doesn't always detect chapter change detection

## Steps to reproduce
1. Go to https://terraso-staging.mystagingwebsite.com/embed-test/
`<iframe src="https://app.terraso.org/tools/story-maps/8c0df6eb/the-mud-that-makes-us/embed" title="Terraso Story Map" height="500px" width="100%"></iframe>`

2. Scroll down
3. See issue
 
## Expected Behavior
Map view changes from chapter to chapter
Chapter opacity is 100% when viewer looks at it

## Actual behavior
Map doesn't move between chapters
Chapter is transparent

Image

## Additional info 
This issue doesn't occur in full-screen mode, or in pinned map mobile experience
https://app.terraso.org/tools/story-maps/8c0df6eb/the-mud-that-makes-us

**Embed view**

**Mobile view**

garobrik:
## Description
Fixes chapter scroll detection so it works in all cases (embedded, preview, editor, view). Includes some minor refactoring of StoryMap code. Main change is to remove the `scrollama` library, and replace with a ~60 line custom hook (`useActiveStep.ts`). No new tests because we were mocking `scrollama` in the first place, this bug could only have been detected by running tests in a proper browser which we don't do.

### Checklist
- [ ] New tests added

### Related Issues
Fixes #2983

### Verification steps
Story maps should detect scroll position correctly now!


src/storyMap/components/StoryMapForm/index.jsx
```diff
@@ -274,7 +274,7 @@ const StoryMapForm = props => {
           onDelete={onDeleteChapter}
           onMoveChapter={onMoveChapter}
         />
-        <Box sx={{ flex: 1, overflowY: 'scroll', position: 'relative' }}>
```
garobrik, line 277: turns out these weren't doing anything anymore 🤷

src/storyMap/components/StoryMapForm/ChapterForm.jsx
```diff
@@ -231,15 +232,15 @@ const ChapterForm = ({ theme, record }) => {
...
       className={classList}
       direction="row"
       component="section"
       aria-label={t('storyMap.view_chapter_label', {
         title: record.title || t('storyMap.form_chapter_no_title_label'),
       })}
+      sx={{ opacity: 0.99 }}
     >
-      {/* div with ID added because of an Intersection Observer issue with overflow */}
-      <div className="step" id={record.id}></div>
```
garobrik, line 242: we were using a different approach here for the IntersectionObserver target (0 height div instead of targeting the whole chapter element), this is unified now, but let me know if there's some reason to keep this

src/storyMap/components/StoryMap.css
```diff
@@ -41,27 +41,6 @@ footer {
...
-}
-#map-inset .mapboxgl-ctrl-bottom-left {
-  display: none;
-}
-@media (max-width: 500px) {
-  #map-inset {
-    display: none;
-  }
-}
-
```
garobrik, line 64: this should've been removed during area of interest changes but i forgot, just a little throw-in here

src/storyMap/components/StoryMap.css
```diff
@@ -129,12 +108,8 @@ footer {
 .step-container {
   padding-top: 15vh;
   padding-bottom: 35vh;
-  opacity: 0.25;
   display: flex;
 }
-.step-container.active {
-  opacity: 0.99;
-}
```
garobrik, line 137: all these CSS changes just got moved to JSX. idk what the logic is for how much to move, but i'm hoping we can eventually remove this file since it's a bit weird that it's the only place we style with CSS instead of our standard MUI stuff. maybe during component library work?

src/storyMap/components/StoryMap.jsx
```diff
@@ -352,14 +319,20 @@ const StoryMap = props => {
...
         className={ALIGNMENTS[config.alignment]}
       >
-        <Scroller onStepChange={onStepChangeWrapped} onReady={onReady} />
-        <TitleComponent config={config} />
-        {filteredChapters.map(chapter => (
-          <ChapterComponent
-            key={chapter.id}
-            theme={config.theme}
-            record={chapter}
+        <div ref={registerStep} id={STORY_MAP_TITLE_ID}>
```
josebui, line 322: Potential issue: It seems that this id `STORY_MAP_TITLE_ID` is also used in the `TitleForm` component

josebui (approved):
Looks good, just one small comment

PR #2996, opened 2026-04-28
feat: StoryMap buffering performance

josebui:
## Description

Story map performance improvements

### Checklist
- [x] New tests added


src/storyMap/components/StoryMapForm/EditableText.jsx
```diff
@@ -29,6 +29,7 @@ const EditableText = props => {
     focus,
     label,
   } = props;
+  const normalizedValue = value || '';
```
garobrik: question: there's still a couple places below where we use `value` instead of `normalizedValue`, should we just use `normalizedValue` for everything?

src/storyMap/components/StoryMapForm/ChapterForm.jsx
```diff
@@ -157,45 +156,31 @@ const ChapterConfig = props => {
   );
 };
 
-const ChapterForm = ({ theme, record }) => {
+const ChapterForm = props => {
```
garobrik, line 163: question: i don't really understand why we're using `persistedChapter` in some places and `displayedChapter` in other places in this function, it seems like data from one or the other should be shown to the user, but never a mixture?

src/storyMap/components/StoryMapForm/index.jsx
```diff
@@ -102,44 +116,69 @@ const StoryMapForm = props => {
...
+    [flushBufferedChapterEdits, markRevisionSaved, mediaFiles]
+  );
+
+  const savePersistedDraft = useCallback(
+    (mediaFilesToSave, revision) =>
+      onSaveDraft(getConfig(), mediaFilesToSave, revision).then(() => {
+        markRevisionSaved(revision);
+      }),
+    [getConfig, markRevisionSaved, onSaveDraft]
+  );
```
garobrik: question: how do we know the config from `getConfig()` and the `revision` match up?

src/storyMap/components/StoryMapForm/index.jsx
```diff
@@ -49,6 +49,20 @@ const BASE_CHAPTER = {
...
+  saving,
+  saveError,
+}) => ({
+  configRevision,
+  mediaFiles,
+  isConfigDirty,
+  saving,
+  saveError,
+});
+
```
garobrik: question: why do we need a defensive copy here?

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
```diff
@@ -24,17 +24,69 @@ import {
...
   const [config, setConfig] = useState(baseConfig || {});
+  const [configRevision, setConfigRevision] = useState(0);
   const [preview, setPreview] = useState(false);
   const [mediaFiles, setMediaFiles] = useState({});
-  const [isDirty, setIsDirty] = useState(false);
+  const [isConfigDirty, setIsConfigDirty] = useState(false);
+  const [hasBufferedChapterChanges, setHasBufferedChapterChanges] =
+    useState(false);
   const init = useRef(false);
+  const latestConfigRef = useRef(baseConfig || {});
```
garobrik: question: why would we ever have a falsy `baseConfig` here, and why would it be desirable/valid to have an empty object has the `latestConfig`. i'd worry about a subtle bug unintentionally passing `undefined` as `baseConfig` from above and then persisting the empty config to the backend. it seems like we should throw an error or do something else if we create a config context without a valid config?

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
```diff
@@ -48,68 +100,203 @@ export const StoryMapConfigContextProvider = props => {
...
   const getMediaFile = useCallback(id => mediaFiles[id]?.content, [mediaFiles]);
 
-  const saved = useCallback(() => setIsDirty(false), []);
+  const isCurrentRevision = useCallback(
+    revision =>
+      revision === undefined ||
+      revision === null ||
+      revision === configRevisionRef.current,
+    []
+  );
```
garobrik, line 133: question: why do undefined or null revisions count as "current"?

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
```diff
@@ -48,68 +100,203 @@ export const StoryMapConfigContextProvider = props => {
...
+  const applySavedRevisionConfig = useCallback(
+    (revision, savedConfig) => {
+      if (!isCurrentRevision(revision)) {
+        return false;
+      }
+
+      const nextConfig = resolveConfigUpdate(
+        latestConfigRef.current,
+        savedConfig
+      );
```
garobrik: question: why do we need to resolve the dataLayerIds for fresh configs we just got from the server?

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
```diff
@@ -24,17 +24,69 @@ import {
...
+    configRevisionRef.current = nextRevision;
+    setConfig(nextConfig);
+    setConfigRevision(nextRevision);
+    setIsConfigDirty(dirty);
+
+    return {
+      config: nextConfig,
+      revision: nextRevision,
+    };
+  }, []);
```
garobrik: question: why would we want to do this but not mark `dirty` as `true` (same question for `updateConfig` and `flushBufferedChapterEdits`)

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
```diff
@@ -48,68 +100,203 @@ export const StoryMapConfigContextProvider = props => {
...
+        return;
+      }
+
+      if (hasBufferedChanges) {
+        chaptersWithBufferedChanges.add(chapterId);
+      } else {
+        chaptersWithBufferedChanges.delete(chapterId);
+      }
+
+      setHasBufferedChapterChanges(chaptersWithBufferedChanges.size > 0);
```
garobrik, line 192: question: why is there no timing issue with using a `setState` for the `hasBufferedChapterChanges` dirty tracker but we're using `ref`s for other dirty state tracking?

src/storyMap/components/StoryMapForm/useBufferedChapterFields.js
```diff
@@ -0,0 +1,379 @@
...
+  const hasPendingBufferedFieldChanges = !_.isEmpty(
+    pendingBufferedFieldChanges
+  );
+
+  useEffect(() => {
+    latestBufferedSnapshotRef.current = {
+      bufferedFieldValues,
+      chapter,
+    };
+  }, [bufferedFieldValues, chapter]);
```
garobrik: question: could we directly set this ref when we update the bufferedFieldValues instead of doing it later in a `useEffect`? otherwise it seems there's a period where the "latest" ref is not actually the latest, although i'm not sure whether that matters timing-wise.

garobrik (approved):
ok, i had a pretty hard time understanding what was going on in this PR, but i didn't find any issues! there's a bunch of places where i left questions.

i think it would be helpful to have a comment explaining the data flow/architecture for both the config revision logic and the buffered chapter logic (which do feel also like separate changes even if they're complementary).

at this point the auto-save logic is getting complicated enough that it makes me wonder if we should be looking into a library, because it feels like a lot of very entangled/complicated code to be maintaining ourselves, especially when it's still the case that a stray frontend bug could delete or corrupt someone's story map permanently on the backend.

DerekCaelin:
> at this point the auto-save logic is getting complicated enough that it makes me wonder if we should be looking into a library, because it feels like a lot of very entangled/complicated code to be maintaining ourselves, especially when it's still the case that a stray frontend bug could delete or corrupt someone's story map permanently on the backend.

Added a tech-debt task: [Research auto-save library](https://github.com/techmatters/terraso-web-client/issues/3005)	https://github.com/techmatters/terraso-web-client/issues/3005 - how high pri do you think it is?

src/storyMap/components/StoryMapForm/index.jsx
josebui: I removed it

src/storyMap/components/StoryMapForm/ChapterForm.jsx
josebui: This was improved

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
josebui: This is fixed

src/storyMap/components/StoryMapForm/EditableText.jsx
josebui: I removed normalizedValue value

src/storyMap/components/StoryMapForm/useBufferedChapterFields.js
josebui: this was updated

src/storyMap/components/StoryMapForm/index.jsx
josebui: This was updated

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
josebui: This was updated

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
josebui: This was updated

src/storyMap/components/StoryMapForm/storyMapConfigContext.jsx
josebui: This was not added by this branch, it was just moved

PR #3021, opened 2026-06-23
feat(story-map): add selectable story map themes

josebui:
### Checklist
- [x] New tests added

### Related Issues
- https://github.com/techmatters/terraso-product/issues/1557

### Verification steps

image

### DIFF
--- a/src/common/components/RichTextEditor/index.jsx
+++ b/src/common/components/RichTextEditor/index.jsx
@@ -758,13 +758,9 @@ const RichTextEditor = props => {
       addContainer
         ? withProps(Paper, {
             variant: 'outlined',
-            sx: [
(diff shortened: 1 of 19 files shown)


garobrik (approved):
LGTM!

out of curiosity, did you implement using CSS variables to keep the change smaller, or do you prefer having some of the story map styling logic in `StoryMap.css`? 

I've been slowly moving things out of there because it's inconsistent with how we style the rest of the app, so if you'd like to keep things in there we should probably talk about it!

josebui:
@garobrik About the styles, yes, that felt like the simplest most idiomatic way, not a strong opinion on where to put it but felt like the best approach for this, we can defenitely talk about improving or changing that

